Issue #24, opened 2025-06-13
Build GUI Chatbox for GPULlama3.java Inference  with embedded GPU resource monitoring


mikepapadim:
## Description:

We want to create a user-friendly chat-style GUI for running LLM inference via our Java engine (powered by TornadoVM), inspired by the layout shown in the attached screenshot.

##  Desired Features

* Prompt Input
   * A text box where users can enter prompts for the LLM

* Dropdown Selectors
   * Engine: Choose between TornadoVM, JVM, etc.
   * Model: Select the model file to use for inference between Llama, Mistral etc

* File Picker
   * A "Browse" button to select .java files or model directories

* Run Button
    * A button to trigger inference (e.g., using tornado-llama-opencv)

* Output Display
   * A read-only area to show model responses and logs
   * Optional: System Monitoring Panel
   * Live GPU usage via nvtop or nvidia-smi
   * Live CPU & memory stats via htop or Java system metrics

Preferably, integrated with the core code base with JavaFX, or external with QT or python QT.

## Example POC:

![Image](https://github.com/user-attachments/assets/ea958a0c-20c9-4045-9fca-0029fce6edb4)

svntax:
I'd like to take on this issue, but does this feature also depend on #17 being merged first?

PR #33, opened 2025-06-25
Add GUI Chatbox for GPULlama3.java Inference

svntax:
This PR adds a new JavaFX GUI for running inference with `GPULlama3` (for issue #24).

It adds a new package `com.example.gui` containing all the new classes for the chatbox GUI, following a Model-View-Controller-Interactor framework.

## Key Features

- Dropdown menu to select an engine (TornadoVM, JVM).
- Browse button to select the directory to the user's `GPULlama3.java` install.
- Dropdown menu and Reload button to search for models inside of a `/models` folder in the user's `GPULlama3.java` directory.
- Prompt text field for user input.
- Run button to trigger inference by running `llama-tornado` as a new process.
- Output area to display responses and other logs.

## How to Run

After following the "Install, Build, and Run" instructions from the README, run the following:

`mvn javafx:run`

## Notes

- Dark theme styling is from AtlantaFX.
- The right panel of the GUI is unfinished, but contains the System Monitoring panel with checkboxes (that do nothing), and an empty text area where the monitoring terminals would display.
- I'm using Windows, so Linux / macOS is untested.

## Next Steps

- I can try to add the system monitoring features, although I'm not sure how far I'll get since I'm on Windows, so I can't test `htop`, `nvtop` or any of the Linux-specific options as far as I know.
- Is the system monitoring display supposed to be embedded terminals? I did a bit of searching and found projects like TerminalFX and JediTermFX for this, but I don't know if that's the best way to implement this.

### DIFF
--- a/src/main/java/com/example/gui/ChatboxInteractor.java
+++ b/src/main/java/com/example/gui/ChatboxInteractor.java
@@ -1,8 +1,12 @@
 package com.example.gui;
 
-import java.io.BufferedReader;
-import java.io.IOException;
-import java.io.InputStreamReader;
+import com.example.LlamaApp;
+import com.example.Options;
+import com.example.inference.sampler.Sampler;
+import com.example.model.Model;
+
+import java.io.ByteArrayOutputStream;
+import java.io.PrintStream;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.List;
@@ -15,28 +19,20 @@ public class ChatboxInteractor {
         this.model = viewModel;
     }
 
-    // Run the 'llama-tornado' script in the given Llama3 path, and stream its output to the GUI's output area.
-    public void runLlamaTornado() {
+    private String[] buildCommands() {
         List<String> commands = new ArrayList<>();
-
-        // Format for running 'llama-tornado' depends on the operating system.
-        if (System.getProperty("os.name").startsWith("Windows")) {
-            commands.add("external\\tornadovm\\.venv\\Scripts\\python");
-            commands.add("llama-tornado");
-        } else {
-            commands.add("llama-tornado");
-        }
 
         ChatboxModel.Engine engine = model.getSelectedEngine();
         if (engine == ChatboxModel.Engine.TORNADO_VM) {
-            commands.add("--gpu");
+            // TODO: LlamaApp.USE_TORNADOVM is a final constant, but the GUI needs to be able to set this value
+            //commands.add("--gpu");
         }
 
         // Assume that models are found in a /models directory.
         String selectedModel = model.getSelectedModel();
         if (selectedModel == null || selectedModel.isEmpty()) {
             model.setOutputText("Please select a model.");
-            return;
+            return null;
         }
         String modelPath = String.format("./models/%s", selectedModel);
         String prompt = String.format("\"%s\"", model.getPromptText());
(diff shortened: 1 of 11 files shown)


mikepapadim:
Thank you @svntax it worked just fine on my setup!

mikepapadim (commented):
This is great! Thank you for the contribution for #24 

I have some suggestios to make it a bit more robust:

1) Lets avoid the indirection of calling the Python script from the tha Java process:
   Lets extend the main to have something like this and make gui to be part of the actual application: 
   ```java
            // LlamaApp.java
            public static void main(String[] args) throws IOException {
                Options options = Options.parseOptions(args);
            
                if (options.guiMode()) { // Add a new guiMode() option
                    // Launch the JavaFX application
                    Application.launch(LlamaGui.class, args);
                } else {
                    // Run the existing CLI logic
                    Model model = loadModel(options);
                    Sampler sampler = createSampler(model, options);
                    if (options.interactive()) {
                        model.runInteractive(sampler, options);
                    } else {
                        model.runInstructOnce(sampler, options);
                    }
                }
            }
  
   ```
2) then we can modify the `llama-tornado` script to add a --gui flag and launch the gui from there.
3) we need to add to GUI a check box or drop-down menu for the 2 options -> either to run instruct mode or interactive.

The output text then it can be directly obtained from -> 
1) https://github.com/beehive-lab/GPULlama3.java/blob/460be9984ab6727daa2f03c949fa602060893b9c/src/main/java/com/example/model/Model.java#L179
2) https://github.com/beehive-lab/GPULlama3.java/blob/460be9984ab6727daa2f03c949fa602060893b9c/src/main/java/com/example/model/Model.java#L158
3) in interactive case -> https://github.com/beehive-lab/GPULlama3.java/blob/460be9984ab6727daa2f03c949fa602060893b9c/src/main/java/com/example/model/Model.java#L83

if these changes make sense to you feel free to extend the PR.

Thanks

src/main/java/com/example/gui/ChatboxInteractor.java
```diff
@@ -0,0 +1,89 @@
...
+            model.setOutputText("Please set the Llama3 path:");
+            return;
+        }
+
+        // Format for running 'llama-tornado' depends on the operating system.
+        if (System.getProperty("os.name").startsWith("Windows")) {
+            commands.add(String.format("%s\\external\\tornadovm\\.venv\\Scripts\\python", llama3Path));
+            commands.add("llama-tornado");
+        } else {
+            commands.add(String.format("%s/.llama-tornado", llama3Path));
```
mikepapadim:
```suggestion
            commands.add(String.format("%s/llama-tornado", llama3Path));
```

`.llama-tornado` does not work on Linux

src/main/java/com/example/gui/ChatboxViewBuilder.java
```diff
@@ -0,0 +1,245 @@
...
+        });
+
+        TextField pathField = boundTextField(model.llama3PathProperty());
+        HBox box = new HBox(8, createLabel("Llama3 Path:"), pathField, browseButton);
+        box.setAlignment(Pos.CENTER_LEFT);
+        HBox.setHgrow(pathField, Priority.ALWAYS);
+        pathField.setMaxWidth(Double.MAX_VALUE);
+
+        return box;
+    }
```
mikepapadim: This box I think is not required as the GUI is laucnhed from within the application and it is running the ROOT directory of the project. So, no need to specify that.

svntax:
I'll work on these changes soon.

svntax:
I'm a bit stuck on how to run the application directly from the GUI. I have it working for instruct mode, but interactive mode currently reads input from the command line in a loop (in `Model`'s `runInteractive()` method). Would we need a new method or changes to `Model` for interactive mode to work with the GUI? Or is there some other better approach?

Two other things I'd like feedback on:
- To run models from the GUI, I'm using `LlamaApp`'s methods `loadModel()` and `createSampler()` from within `ChatboxInteractor`, so is it okay to make them public?
- Using TornadoVM requires `LlamaApp.USE_TORNADOVM` to read from config flags, but if we want the GUI to let the user choose, how should this change when right now it's a constant?

mikepapadim:
> I'm a bit stuck on how to run the application directly from the GUI. I have it working for instruct mode, but interactive mode currently reads input from the command line in a loop (in `Model`'s `runInteractive()` method). Would we need a new method or changes to `Model` for interactive mode to work with the GUI? Or is there some other better approach?
> 
> Two other things I'd like feedback on:
> 
> * To run models from the GUI, I'm using `LlamaApp`'s methods `loadModel()` and `createSampler()` from within `ChatboxInteractor`, so is it okay to make them public?
> * Using TornadoVM requires `LlamaApp.USE_TORNADOVM` to read from config flags, but if we want the GUI to let the user choose, how should this change when right now it's a constant?


-Handling Interactive Mode with the GUI
Modifying runInteractive() to work with the GUI isn't ideal because its while loop is designed for a command-line interface (CLI).

A cleaner approach is to create a new method in the Model class specifically for the GUI. This method would take a single user input (as a string) and return the model's single response. This way, the GUI can manage the "loop" itself—calling this new method every time the user sends a message.

-Making loadModel() and createSampler() Public
Yes, making loadModel() and createSampler() in LlamaApp public is a perfectly reasonable solution.

The ChatboxInteractor is acting as a controller or intermediary between your GUI and the core application logic in LlamaApp. For the GUI to be able to trigger the model loading and sampler creation process, the methods that perform these actions must be accessible to it. Encapsulation is important, but in this case, you are intentionally exposing specific functionalities to the GUI layer, which is a standard and necessary practice.

-Managing the USE_TORNADOVM Flag
You're right; a static final constant won't work if you want the user to be able to change this setting from the GUI.

The best way to handle this is to change USE_TORNADOVM from a constant to a regular member variable within a configuration object or directly in LlamaApp.

Remove final: Change public static final boolean USE_TORNADOVM to something like private boolean useTornadoVM.

Add a Setter Method: Create a public method to change its value, for example, public void setUseTornadoVM(boolean value).

GUI Integration: Your GUI's checkbox or toggle can now call this setter method before it calls loadModel().

This way, the user's choice is set first, and then the model is loaded with the correct configuration. The setting is no longer a compile-time constant but a runtime configuration option.

svntax:
I finally have the GUI working directly with the main application, but there's a memory leak when using TornadoVM. I'm not sure if there's a problem with the approach I have, or if I'm not freeing up resources correctly, or if it's something else.

## How to Run
Use the new `--gui` flag in llama-tornado to launch the GUI.

Windows example:
```
python llama-tornado --gui
```

## How it works
I copied how LlamaApp runs by creating a Model and Sampler when starting a new chat. Instruct mode uses `runInstructOnce()`, and interactive mode uses the new method in Model `runInteractiveStep()`, which returns a Response object (a record in Model) with data for the ongoing conversation (state, conversation tokens, and a TornadoVMMasterPlan if applicable). These are reused every time the user sends a message in interactive mode, until "quit" or "exit" is sent, which then ends the chat session and should free resources.

## Memory leak problem
I've tested with Llama-3.2-1B-Instruct-Q8_0.gguf, and I can confirm that in both instruct and interactive mode, if using TornadoVM, the model does load correctly (about 3-4 GB to VRAM), but when I try to free resources with `freeTornadoExecutionPlan()`, it frees up around 500 MB only.

Since the inference is happening in a background thread, I thought maybe it has to happen in the main JavaFX thread, but that didn't help.

The problem happens only with TornadoVM. There's no memory leak as far as I can tell if running on CPU (selecting JVM for the engine).

mikepapadim:
@svntax thank you for contributing the GUI, let me try it and see if I can find the root cause of the memory leak

PR #46, opened 2025-09-01
Improve attention performance for qwen2.5 & deepseek

orionpapadakis:
This PR adds a new Tornado kernel for parallel attention and improves grid sizes calculations.


src/main/java/org/beehive/gpullama3/tornadovm/Qwen2TornadoVMLayerPlanner.java
```diff
@@ -190,12 +190,21 @@ private GridScheduler setupQwen2GridSchedulersLayeredNonNvidia() {
         rmsNormWorker.setLocalWork(32, 1, 1);         // Set local work size to 256 (standard efficient size)
 
         // Parallel attention worker configuration
```
mikepapadim, line 192: put it into a seperate method: 
and make 64 constant etc

```java
    public static int computeOptimalLocalSize(int headSize) {
        int optimalLocalSize = Math.min(headSize, 64);

        if (headSize % optimalLocalSize != 0) {
            for (int size = optimalLocalSize; size >= 1; size--) {
                if (headSize % size == 0) {
                    return size;
                }
            }
        }

        return optimalLocalSize;
    }
```

PR #49, opened 2025-09-08
Rest API support 

mikepapadim:


orionpapadakis:
The already supported and actively maintained integration of GPULlama3.java with Quakus-langchain4j provides equivalent functionality (server mode) consequently, we decided to postpone the introduction and active maintenance of a similar/mirror, standalone implementation within the GPULlama3.java codebase for now.

PR #63, opened 2025-11-07
Clean support for Non-Nvidia hardware vendors like Apple Silicon

mikepapadim:


src/main/java/org/beehive/gpullama3/model/loader/ModelLoader.java
```diff
@@ -59,7 +59,7 @@ private static ModelType detectModelType(Map<String, Object> metadata) {
                 return ModelType.QWEN_2;
             } else if (lowerName.contains("qwen3")) {
                 return ModelType.QWEN_3;
-            } else if (lowerName.contains("deepseek r1 distill")) {
+            } else if (lowerName.contains("deepseek")) {
```
orionpapadakis: i think we still need this

src/main/java/org/beehive/gpullama3/tornadovm/layers/AbstractLayer.java
```diff
@@ -18,7 +18,8 @@ public abstract class AbstractLayer {
 
     /** Common constants used in tasks & worker-grid sizing. */
     protected static final int LOCAL_WORK_GROUP_SIZE_ALLOC = 32;
-    protected static final int THREAD_SCALE_FOR_LOGITS = 1;
+    // TODO: 1 OR 8?
+    protected static final int THREAD_SCALE_FOR_LOGITS = 8;
```
orionpapadakis, line 21: remove todo

PR #77, opened 2025-12-01
RMS normalization kernel optimization by fusing the reduction kernel and context mapping kernel

yrq0208:
The main optimization is done by using all the threads, instead of only the first global thread, to calculate the scaling factor. This avoids thread divergence and the need to sync threads across work groups, since in context mapping, all the threads are used instead of just global thread 0 (sync threads across work groups is difficult to implement in OpenCL since it involves the use of atomic operations and locks, though maybe doable in CUDA with grid sync). By doing this, the reduction kernel and the context mapping kernel can be merged, reducing kernel launch overheads.

Tested with the following models: beehive-llama-3.2-1b-instruct-fp16.gguf, beehive-llama-3.2-3b-instruct-fp16.gguf, DeepSeek-R1-Distill-Qwen-1.5B-F16.gguf, Qwen2.5-0.5B-Instruct-f16.gguf, qwen2.5-1.5b-instruct-fp16.gguf, Qwen3-0.6B-f16.gguf, Qwen3-1.7B-f16.gguf, Qwen3-4B-f16.gguf. 7B and 8B models are not tested due to hardware limitations (5080 only has 16GB of VRAM). 

E2E performance (in tok/s) improvement ranges between 5% - 22% on 5080 mobile with TornadoVM OpenCL backend.

### DIFF
--- a/llama-tornado
+++ b/llama-tornado
@@ -410,7 +410,7 @@ def create_parser() -> argparse.ArgumentParser:
         const=Backend.PTX,
         help="Use PTX/CUDA backend",
     )
-    hw_group.add_argument("--gpu-memory", default="7GB", help="GPU memory allocation")
+    hw_group.add_argument("--gpu-memory", default="14GB", help="GPU memory allocation")
     hw_group.add_argument("--heap-min", default="20g", help="Minimum JVM heap size")
     hw_group.add_argument("--heap-max", default="20g", help="Maximum JVM heap size")
 
--- a/llama-tornado
+++ b/llama-tornado
@@ -498,6 +498,11 @@ def main():
     parser = create_parser()
     args = parser.parse_args()
 
+    # Set default profiler log path relative to LLAMA_ROOT
+    if args.profiler_dump_dir is None:
+        llama_root = os.environ.get("LLAMA_ROOT")
+        args.profiler_dump_dir = os.path.join(llama_root, "profiler-log.json")
+
     # Set default seed if not provided
     if args.seed is None:
         args.seed = int(time.time())
--- a/llama-tornado
+++ b/llama-tornado
@@ -422,7 +422,7 @@ def create_parser() -> argparse.ArgumentParser:
     )
     debug_group.add_argument(
         "--profiler-dump-dir",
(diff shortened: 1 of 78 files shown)


yrq0208:
Looks like most of the CI checks are fine apart from Phi-3-mini-4k-instruct-fp16.gguf running with PTX? Though this seems to be an ongoing issue with the baseline (https://github.com/beehive-lab/GPULlama3.java/actions/runs/19747335165/job/56616800315?pr=75)?

set_paths
```diff
@@ -6,15 +6,19 @@
...
 # Add TornadoVM and LLaMA bin directories to PATH
-export PATH="${PATH}:${TORNADO_SDK}/bin:${LLAMA_ROOT}"
+export PATH="${PATH}:${TORNADO_SDK}:${LLAMA_ROOT}"
 
 # Optional: Set JAVA_HOME if required
 # export JAVA_HOME=/path/to/graalvm
 # export PATH="${JAVA_HOME}/bin:${PATH}"
 
-echo "[INFO] Environment configured for LLaMA3 with TornadoVM at: $TORNADO_SDK"
+echo "[INFO] Environment configured for LLaMA3 with TornadoVM at: $TORNADO_ROOT"
```
stratika: @yrq0208 can you please revert these changes and try it with the TORNADO_SDK instructions in readme?

yrq0208: done

stratika (commented):
Can you remove the external/tornadovm path? This directory is removed.

yrq0208:
> Can you remove the external/tornadovm path? This directory is removed.

done.

llama-tornado
```diff
@@ -498,11 +498,6 @@ def main():
     parser = create_parser()
     args = parser.parse_args()
 
-    # Set default profiler log path relative to LLAMA_ROOT
```
stratika, line 501: Any specific reason, why we removed the profiler option in this PR?

yrq0208: reverted

stratika (approved):
LGTM

stratika:
Thank you, we decided to postpone this for later due to instability with some models. Feel free to open a new PR and provide coverage with models.

PR #102, opened 2026-04-02
Add Prefill–Decode Separation with Batched Prompt Ingestion and Logits Skipping 

orionpapadakis:
This PR implements the prefill-decode concept in GPULlama3.java:

1) **Prefill** (or prompt ingestion) is the inference pass for prompt (input) tokens. Currently they are ingested sequentially (one-by-one). However, all prompt tokens are already known and also they are independent entities; hence the sequential ingestion is sub-optimal. A well known practice is to ingest prompt tokens in batches (i.e., of 32) instead of one-by-one. In addition, logits creation can be skipped for all prompt tokens but the last one as they are not used at all. The logits of the last prompt token are used to get the first generated token which in its turn feeds the decode phase.

2) **Decode** (new token generation) is the inference pass for each generated token. In contrast to prefill, this phase remains token-sequential (as it is now) because each generated token depends on the previous one.

Based on the above, this PR breaks down the prefill-decode feature implementation in 4 discrete phases:

- **Phase 1**: CPU prefill/decode split — sequential, skip logits during prefill                                                                                                                               

- **Phase 2**: GPU prefill/decode split — sequential, skip logits during prefill

- **Phase 3**: CPU batched prefill (batch size B, default 32)                                                                                                                                              

- **Phase 4**: GPU batched prefill (batch size B, default 32)

---                                                   
                                                                                                                                                                                                                    
  ## Implementation                                     

  ### Top-level dispatch

  `Llama#generateTokens` / `#generateTokensGPU` perform a three-way dispatch:

  PREFILL_BATCH_SIZE > 1  →  InferenceEngineWithBatchPrefillDecode   (Phase 3/4)                                                                                                                                    
  PREFILL_BATCH_SIZE == 1 →  InferenceEngineWithPrefillDecode         (Phase 1/2)
  (default)               →  InferenceEngine                          (standard)                                                                                                                                    
                                                                                                                                                                                                                    
  ### Per-phase entry points
                                                                                                                                                                                                                    
  | Phase | Engine | Core primitive | Execution plan |                                                                                                                                                              
  |---|---|---|---|
  | 1 — CPU sequential | `InferenceEngineWithPrefillDecode#generateTokensLlama` | `InferenceCoreWithPrefillDecode#forwardJavaPrefill` | — |                                                                         
[...]

PR #106, opened 2026-04-12
Detailed Performance Metrics (Resolves Issue #104)

### LINKED ISSUE #104: Improve performance metrics collection
It would be very helpful to expose more detailed performance metrics in gpullama, similar to what Ollama provides ([see](https://docs.ollama.com/api/usage)) .

Right now, it is difficult to properly evaluate performance (especially across CPU/GPU backends and TornadoVM execution) without fine-grained timing information. Having a consistent set of metrics would significantly improve benchmarking, profiling, and optimization.

##### Core metrics (aligned with Ollama-style reporting):

`total_duration` – total time to generate the full response
`load_duration` – time spent loading the model
`prompt_eval_count` – number of input tokens processed
`prompt_eval_duration` (prefill) – time spent processing the prompt
`eval_count` – number of generated output tokens
`eval_duration` (decode) – time spent generating tokens

##### TornadoVM-specific metrics:

`tornado_task_graph_compile_duration` – time to compile the Tornado task graph
`tornado_task_graph_warmup_duration` – time spent in warmup/execution until steady state

All timings should ideally be reported in nanoseconds for consistency and precision.

With the above we can calculate:
- `time_to_first_token` (TTFT) – can be derived from existing durations (e.g. load + prefill + first decode step), so it may not need separate instrumentation if timestamps are available
- `prefill_throuput` as `tok/s` = `prompt_eval_count` / `prompt_eval_duration`
- `decode_throuput` as `tok/s` = `eval_count` / `eval_duration`
- `total_throuput` as `tok/s` = `prompt_eval_count` + `eval_count` / `total_duration` <--- as we do now

#### Why this is useful

These metrics would make it easier to:

break down execution into loading, prefill, decode, and runtime overheads
understand TornadoVM-specific costs (compilation and warmup)
compare CPU vs GPU vs TornadoVM performance more accurately
identify bottlenecks and guide optimizations

LegL0ngly:
Resolves #104 

Added functionality to keep track of the fine-grained performance metrics described in Issue #104.

In LastRunMetrics:

- Modified the record to accept all necessary variables (except for model load duration) as parameters and updated the record's documentation accordingly.
- Added a static long for load duration, because it will not change between every print. Also added a method for setting the load duration to be called from ModelLoader.
- Modified the existing print statement to print new parameters, and print durations in both nanoseconds and seconds. Also created if statements to print the model load duration only once, and the TornadoVM metrics only when using the GPU.

In InferenceEngine:

- Sent new metric information to the setMetrics method in LastRunMetrics
- TornadoVM metrics are retrieved from TornadoVMMasterPlan when using GPU, otherwise they are set to 0, so they are not printed.
- Model load time is NOT retrieved here; the load duration is sent to LastRunMetrics from ModelLoader because it does not need to be retrieved after every response.
- Added a timer for inference start in CPU Phi3, which was previously missing
- Changed the name of the TornadoVMMasterPlan object in the GPU Granite method to be consistent with the naming of all other TornadoVMMasterPlan objects in InferenceEngine.

In ModelLoader:

- Added simple model duration timers and logic; times sent to LastRunMetrics
- Declared a new model instead of returning the result of modelType.loadModel immediately to time the loading duration properly.

In TornadoVMMasterPlan:

- Added variables to keep track of compile and warmup duration, as well as corresponding accessor and mutator methods. The accessor methods are called from InferenceEngine and passed to the setMetrics method in LastRunMetrics. The mutators are called when times are calculated within TornadaVMMasterPlan. 
- Timing variables were already in place, but only written to when using a flag for debugging. They are now always written to, and the if statements for the flags are still there, and will still print the same information if the flag is used.

One last note not related to the changes:
In the contributing instructions, contributors are instructed to PR the "develop" branch. It appears the branch no longer exists, and the instructions are outdated. Might be confusing for new contributors.

### DIFF
--- a/src/main/java/org/beehive/gpullama3/auxiliary/LastRunMetrics.java
+++ b/src/main/java/org/beehive/gpullama3/auxiliary/LastRunMetrics.java
@@ -3,9 +3,9 @@ package org.beehive.gpullama3.auxiliary;
 /**
  * Record to store metrics from the last model run.
  * @param totalTokens The total number of tokens processed
- * @param totalSeconds The total time in seconds
+ * @param totalNanos The total time in nanoseconds
  */
-public record LastRunMetrics(int totalTokens, double totalSeconds) {
+public record LastRunMetrics(int totalTokens, long totalNanos, int promptEvalCount, long promptNanos, int inferenceEvalCount, long inferenceNanos) {
     /**
      * Singleton instance to store the latest metrics
      */
@@ -14,11 +14,11 @@ public record LastRunMetrics(int totalTokens, double totalSeconds) {
     /**
      * Sets the metrics for the latest run
      *
-     * @param tokens The total number of tokens processed
-     * @param seconds The total time in seconds
+     * @param totalTokens The total number of tokens processed
+     * @param totalNanos The total time in nanoseconds
      */
-    public static void setMetrics(int tokens, double seconds) {
-        latestMetrics = new LastRunMetrics(tokens, seconds);
+    public static void setMetrics(int totalTokens, long totalNanos, int promptEvalCount, long promptNanos, int inferenceEvalCount, long inferenceNanos) {
+        latestMetrics = new LastRunMetrics(totalTokens, totalNanos, promptEvalCount, promptNanos, inferenceEvalCount, inferenceNanos);
     }
 
     /**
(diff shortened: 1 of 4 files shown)


mikepapadim:
@LegL0ngly thank you for your contribution, but this feature it was intend to be on top the changes that we are adding in the PR: https://github.com/beehive-lab/GPULlama3.java/pull/102

LegL0ngly:
Ah well, I enjoyed working on it anyway. Thanks for getting back to me quickly.

PR #115, opened 2026-05-12
Several fixes and improvements for CI

orionpapadakis:
This PR:
  
- Merges `integration-quarkus-langchain4j` into `build-and-run` — single workflow with four sequential jobs (`build` → `standalone-inference` / `quarkus-integration` → `perf-history`), eliminating duplicate TornadoVM builds from two concurrent workflows racing on the same runner
- Fix TornadoVM rebuilding on every run — move `TORNADO_ROOT` to `runner.tool_cache` (outside workspace) so actions/checkout's git clean no longer wipes it; SHA-based sentinel skips rebuilds when upstream HEAD is unchanged
- `max-parallel`: 1 on build matrix — opencl and ptx variants run sequentially, no workspace conflicts
- Extract `setup-java` composite action — SDKMAN-based JDK setup deduplicated across all workflows (`build-and-run`, `deploy-maven-central`, `prepare-release`)
- Extract `setup-tornadovm` composite action — TornadoVM clone/build/sentinel logic extracted and reused; find dist deduplication via step output
- Extract run-inference composite action — inference invocation deduplicated across all model steps


stratika:
Thanks @orionpapadakis. LGTM, there is a test run of the pipeline [here](https://github.com/beehive-lab/GPULlama3.java/actions/runs/26404535804).

One note (maybe for future addition) is to provide coverage for different quantization level (f16, q8, etc.). This would be useful, in my opinion.

stratika (approved):

PR #122, opened 2026-06-11
Add prefill-decode and batch-prefill-decode for Qwen3 (FP16 and Q8_0)

orionpapadakis:
## Summary

This PR extends the `prefill-decode` and `batch-prefill-decode` GPU inference paths to Qwen3 models for both FP16 and Q8_0 quantizations. It also adds corresponding coverage in CI.

The core challenge is that Qwen3 diverges from Llama in two ways that affect every attention kernel:

  - **Per-head QK RMSNorm** — Qwen3 applies a separate RMSNorm to each query and key head before RoPE. This requires a dedicated parallel reduction kernel (`Qwen3Kernels.rmsnormReductionWithParallelOffset` / 
  `rmsnormNormalisationWithParallelOffset`) and cannot reuse the scalar Llama path.
  - **GQA layout mismatch** — `Qwen3Configuration.headSize()`, `kvDim()`, and `kvMul()` throw for Qwen3 (dimensions are stored differently). All new Qwen3 layers derive head dimensions from `nEmbdHeadK` / 
  `nEmbdHeadV` / `nHeadKv` directly.
  
  ### State changes

  `State.qDim` and `State.kvDim` are made model-agnostic (previously carried Llama assumptions) so batch-prefill activations can be correctly sized for Qwen3's GQA layout. `LlamaState` and `LlamaConfiguration`
  references in batch-prefill/batch-decode activations are replaced with the base `State` and `Configuration` types.

  ## Verification

  ```bash
  export MODEL=~/LLMModels/Qwen3-0.6B-Q8_0.gguf      # or Qwen3-4B-f16.gguf
  export PROMPT="Explain Newton's second law"

  # Single-token (unchanged baseline)
  ./llama-tornado --gpu --ptx --model $MODEL --prompt "$PROMPT" --max-tokens 256

  # Prefill-decode
  ./llama-tornado --gpu --ptx --model $MODEL --prompt "$PROMPT" --max-tokens 256 \
    --with-prefill-decode

  # Batch-prefill-decode
  ./llama-tornado --gpu --ptx --model $MODEL --prompt "$PROMPT" --max-tokens 256 \
    --with-prefill-decode --batch-prefill-size 32

  # Batch-prefill-decode + CUDA graphs (PTX only)
  ./llama-tornado --gpu --ptx --model $MODEL --prompt "$PROMPT" --max-tokens 256 \
    --with-prefill-decode --batch-prefill-size 32 --cuda-graphs

  All existing Llama FP16 and Q8_0 paths (single-token, prefill-decode, batch-prefill-decode, CUDA graphs) are unaffected.

  CI

  Added CI steps for all four configurations × two quantizations, mirroring the existing Llama coverage:
  - FP16 / Q8_0 — prefill-decode and batch-prefill-decode (both backends)
  - PTX — prefill-decode-cuda-graphs and batch-prefill-decode-cuda-graphs

### DIFF
--- a/src/main/java/org/beehive/gpullama3/tornadovm/layers/type/q8_0/prefill/Qwen3Q8_0LayersBatchPrefill.java
+++ b/src/main/java/org/beehive/gpullama3/tornadovm/layers/type/q8_0/prefill/Qwen3Q8_0LayersBatchPrefill.java
@@ -1,0 +1,250 @@
+package org.beehive.gpullama3.tornadovm.layers.type.q8_0.prefill;
+
+import org.beehive.gpullama3.inference.state.Qwen3State;
+import org.beehive.gpullama3.inference.weights.tornado.Qwen3TornadoWeights;
+import org.beehive.gpullama3.model.qwen3.Qwen3Configuration;
+import org.beehive.gpullama3.tornadovm.kernels.Qwen3Kernels;
+import org.beehive.gpullama3.tornadovm.kernels.TransformerBatchPrefillKernels;
+import org.beehive.gpullama3.tornadovm.scheduling.WorkerGridFactory;
+import org.beehive.gpullama3.tornadovm.layers.BatchPrefillTransformerLayerTaskGraphs;
+import uk.ac.manchester.tornado.api.GridScheduler;
+import uk.ac.manchester.tornado.api.ImmutableTaskGraph;
+import uk.ac.manchester.tornado.api.KernelContext;
+import uk.ac.manchester.tornado.api.TaskGraph;
+import uk.ac.manchester.tornado.api.WorkerGrid;
+import uk.ac.manchester.tornado.api.enums.DataTransferMode;
+
+import java.util.List;
+import java.util.stream.IntStream;
+
+/**
+ * Batched-prefill transformer-layer TaskGraphs for the Qwen3 Q8_0 unified batched prefill-decode plan.
+ *
+ * <p>Q8_0 path: wrapXbBatch (FP32) holds normalized activations; wrapXbFP16Batch is not used.
+ * Mirrors {@link Qwen3FP16LayersBatchPrefill} but uses Q8_0 weights (ByteArray) and FP32
+ * attention normalization path.</p>
+ */
+public class Qwen3Q8_0LayersBatchPrefill implements BatchPrefillTransformerLayerTaskGraphs {
+
+    static final int LOCAL_WORK_GROUP_SIZE = 32;
+
+    private final Qwen3State state;
+    private final Qwen3TornadoWeights weights;
+    private final Qwen3Configuration config;
+    private final KernelContext context = new KernelContext();
+    private final int batchSize;
+    private final int nHeadKv;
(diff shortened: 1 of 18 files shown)


src/main/java/org/beehive/gpullama3/tornadovm/layers/type/q8_0/prefill/Qwen3Q8_0LayersBatchPrefill.java
```diff
@@ -0,0 +1,250 @@
...
+        return layer;
+    }
+    // @formatter:on
+
+    public void updateGridScheduler(GridScheduler scheduler) {
+        int dim    = config.dim();
+        int hidDim = config.hiddenDim();
+
+        WorkerGrid rmsWorker        = WorkerGridFactory.genericWorker(batchSize, 1);
+        WorkerGrid rmsApplyWorker   = WorkerGridFactory.genericWorker(batchSize * dim, 256);
```
stratika, line 199: does it pass code formatting?

src/main/java/org/beehive/gpullama3/tornadovm/plan/ForwardPlanFactory.java
```diff
@@ -176,15 +176,21 @@ private static ForwardPlan createQwen2Q8_0Plan(ExecutionMode mode, Qwen2State st
...
 
     private static ForwardPlan createQwen3FP16Plan(ExecutionMode mode, Qwen3State state, Model model) {
-        if (mode != ExecutionMode.STANDARD)
-            throw new UnsupportedOperationException(mode + " not yet supported for QWEN_3 + F16");
-        return new SingleTokenForwardPlan(model, new Qwen3FP16PlanComponents(state, model));
+        BatchPrefillDecodeForwardPlanComponents components = new Qwen3FP16PlanComponents(state, model);
+        return switch (mode) {
+            case STANDARD             -> new SingleTokenForwardPlan(model, components);
+            case PREFILL_DECODE       -> new PrefillDecodeForwardPlan(model, components);
+            case BATCH_PREFILL_DECODE -> new BatchPrefillDecodeForwardPlan(model, components, TornadoVMMasterPlan.PREFILL_BATCH_SIZE);
```
stratika, line 183: which one is the default mode? the `STANDARD`?

stratika (approved):
LGTM, just minor comments/questions. I tested it on macOS and Linux (with PTX).

src/main/java/org/beehive/gpullama3/tornadovm/plan/ForwardPlanFactory.java
orionpapadakis: yes, standard (single-token) is the current default for two reasons:
1) prefill-decode and batch-prefill-decode is not implemented yet for all supported models
2) batch-prefill-decode currently demands more GPU memory than standard which might not be available to all systems.
Future PRs should work on both above "issues" and finally make the batch-prefill-decode the default execution mode in GPULlama3.java

mikepapadim (approved):

PR #124, opened 2026-06-20
Add pure CUDA backend  along the PTX engine 

mikepapadim:
## Summary

Adds a first-class **`--cuda`** backend path to the `llama-tornado` launcher, mapping to TornadoVM's new **CUDA backend** (CUDA C → NVRTC → PTX → CUDA Driver API). It complements the existing `--opencl`, `--ptx`, and `--metal` paths.

```bash
./llama-tornado --gpu --cuda --model <model.gguf> --prompt "hello"
```

## How it maps to TornadoVM

Backend selection in the launcher is by which TornadoVM driver module is loaded. The new `--cuda` branch mirrors `--ptx` exactly:
- export list `@$TORNADOVM_HOME/etc/exportLists/cuda-exports`
- `--add-modules ...,tornado.drivers.common,tornado.drivers.cuda`

`--gpu --cuda` behaves like `--gpu --ptx`. The `--ptx` help text was tightened (it previously said "PTX/CUDA") now that CUDA is its own flag.

## TornadoVM requirement

The CUDA backend is **not yet in a released TornadoVM**; it lives in **TornadoVM PR #861** — https://github.com/beehive-lab/TornadoVM/pull/861. This PR therefore builds against TornadoVM **`4.0.2-jdk21-dev`** (a build that includes the CUDA backend). The project's own version is unchanged. The README documents this requirement.

## Validation

Built with JDK 21 against TornadoVM `4.0.2-jdk21-dev` and run on an NVIDIA RTX 3070 (device `0:0`, `Backend: CUDA` confirmed via `--print-threads`). All produced coherent output:

| Model | Result |
|---|---|
| `llama-3.2-1b-instruct-q8_0` (Llama Q8) | ✅ "The capital of France is Paris." |
| `Llama-3.2-1B-Instruct.FP16` (Llama FP16) | ✅ coherent |
| `granite-3.2-2b-instruct-Q8_0` (Granite) | ✅ coherent |
| `qwen2.5-1.5b-instruct-q8_0` (Qwen) | ✅ coherent |

No regression: `--opencl`, `--ptx`, and `--metal` still parse and wire their respective driver modules/export lists.

## Changes
- `llama-tornado`: add `CUDA` to the `Backend` enum, add the `--cuda` argparse flag, add the CUDA module-config branch, update the docstring.
- `pom.xml`: build the JDK21 path against TornadoVM `4.0.2-jdk21-dev`.
- `README.md`: list CUDA among supported backends, add a `--cuda` example, document the PR #861 requirement.

PR #127, opened 2026-07-07
Add TornadoVM CUDA backend support and tensor-core (MMA) accelerated batch prefill (FP16 & Q8_0, Llama & Qwen3)

mairooni:
This PR enables GPULlama3.java to run on TornadoVM's new CUDA backend and adds a tensor-core (MMA) accelerated batched-prefill pipeline on top of it. It incorporates and supersedes #124 (this branch is based on it, the CUDA backend changes are carried here so backend support and the MMA path that requires it land together).

All five prefill projections (QKV, Wo, W1/W3, W2) are executed as 128×128-tile GEMMs on tensor cores (m16n8k16, FP32 accumulate) via TornadoVM's MMA kernel API, replacing the per-token matvec formulation. Key elements:

- **Fused projections:** Q/K/V in a single launch writing a packed `[q|k|v]` buffer (fixes grid starvation on the skinny GQA projections), and W1/W3 fused into a packed `[gate|up]` buffer.
- **Software-pipelined GEMMs:** per-thread register staging of the next K-step's tiles overlaps global loads with tensor-core compute.
- **Q8_0 as W8A16:** quantized weights stay in memory and are dequantized to FP16 in the GEMM staging registers — roughly half the weight traffic of FP16 with the same tensor-core pipeline.
- **Rewritten flash attention:** register-partitioned P·V accumulation (removes redundant per-thread work), coalesced K/V tile loads, direct FP16 output.
- **Parallel RMS reductions** with fused residual adds; packed RoPE/SwiGLU; 15 → 12 kernels per layer (13 for Qwen3, including per-head QK-norm).
- **Model coverage:** Llama 3.x and Qwen3 (split-half RoPE, QK-norm, qDim ≠ dim handled), FP16 and Q8_0.
- **Arbitrary batch sizes via M-padding:** The GEMMs operate on fixed 128-row tiles and the MMA API has no predicated store to mask partial tiles, so activation buffers are padded to the next 128-row multiple; padded rows are computed but never read (each output row depends only on its own input row, and non-GEMM kernels use the true batch size). Multiples of 128 give full GEMM efficiency; other sizes work correctly at proportionally reduced efficiency, reported at startup. This is the same tail-tile strategy that dense GEMM libraries use internally (cuBLAS computes full tiles for ragged M too) just visible at the allocation level here.
- **Graceful fallback:** on backends without MMA support, the batch-prefill planners automatically fall back to the portable matvec pipeline

Example usage: 
`./llama-tornado --gpu --cuda --model Q8_0-Llama-3.2-1B-Instruct.gguf --prompt "$LONG_PROMPT" --max-tokens 2048 --with-prefill-decode --batch-prefill-size 128`


mikepapadim (approved):

mairooni:
This PR now passes the workflow. It is ready to be merged.

stratika (approved):

PR #129, opened 2026-07-11
Static batched decode (LLaMA + Qwen3): B independent sequences per step, up to 41x aggregate throughput

mikepapadim:
## Static batched decode — B independent LLaMA sequences per step

Decode **B independent sequences at once**, one token per step, each attending its
**own** KV cache. Turns the bandwidth-bound single-token matvecs of decode into
compute-bound tensor-core GEMMs (one weight read amortized across B tokens), so
aggregate throughput scales ~linearly with B until the 128-row MMA tile fills.

**41× aggregate throughput** vs single-stream decode on an RTX 4090 (Llama-3.2-1B
FP16), output verified coherent + bit-exact against the single-stream greedy reference.

Full write-up in [`BATCHED_DECODE.md`](BATCHED_DECODE.md).

### Why it wins
Single-token decode is memory-bound: each projection reads the whole weight matrix to
make one output vector (intensity ~1). Batch B tokens sharing the weights → each weight
row is read **once** and applied to **all B** → intensity ~B, a GEMM. Past B≈16 the
projections go compute-bound and the tensor cores engage. Attention doesn't batch this
way (per-sequence KV) but is a small fraction of decode FLOPs.

```
             single-token decode              static batched decode (B)
  weight W    read once  ─┐                     read once ─┐
  token x0    ───────────┼─> y0                 ──────────┼─> y0
  token x1    (read W again next step)          ──────────┼─> y1   one GEMM,
  token x2    (read W again) ...                ──────────┼─> y2   W read once
  intensity ~1 (HBM-bound)                      intensity ~B (compute-bound)
```

### Design

One step routine serves **both** phases. Prompt tokens are fed with logits discarded —
fills every slot's KV region with the *same* RoPE the decode step uses (no prefill/decode
mismatch). Then generated tokens feed back.

```
 embXBatch(FP16 B×dim)
      │  [0] prefillActivation      FP16 → FP32  wrapXBatch
      ▼
 ┌──────────  [1..N] batchDecodeLayer_i  (12 tasks, all GEMMs on MMA) ──────────┐
 │ batch_attn_rms → rms_apply → qkvProj(MMA) → batch_rope_kv* → batch_attention*│
 │ → woProj(MMA) → batch_ffn_rms → rms_apply → gateUpProj(MMA) → swiglu         │
 │ → w2Proj(MMA) → w2Resid            (*) only decode-specific kernels           │
 └─────────────────────────────────────────────────────────────────────────────┘
[...]

PR #132, opened 2026-07-14
Fix Qwen3 FP16 garbage on CUDA backend: cross-workgroup race in RMS-norm reduction

mikepapadim:
## Fix: Qwen3 FP16 emits garbage on the CUDA backend — cross-workgroup data race in the RMS-norm reduction

**Symptom.** Qwen3-1.7B FP16 single-token decode on the CUDA backend produces garbage (endless `\n` + stray tokens, argmax stuck) from the first sampled token; Qwen3-4B works. Deterministic, both CUDA-graph modes.

**Root cause.** `reductionOneBlockWithLayer` writes each workgroup's partial sum to `output[groupId+1]`, then thread `gid==0` (workgroup 0) sums `output[1..N]` **in the same kernel with no inter-workgroup synchronization or memory fence** — a data race. On the NON_NVIDIA scheduler this race doesn't exist (a separate `reductionFinalNormalization` task combines the partials); the NVIDIA path skips that task and relies on the racy in-kernel combine. Whether the race is lost is schedule/compilation dependent: Llama-1B and Qwen3-4B happen to win it, Qwen3-1.7B deterministically loses it — every layer's RMS-norm scale is computed from stale partials, corrupting the hidden state everywhere at once.

**Diagnosis** (bisection, all on RTX 4090, TornadoVM 5.0.1-jdk21-dev CUDA backend):
- Batched-decode Qwen3 path (PR #129 kernels) = coherent with the same weights → weights/tokenizer/logits fine.
- Batch-prefill mode produced the correct first token (`<think>`), then the decode graph degraded → single-token layer graph implicated.
- Swapped every Qwen3-specific fused kernel (QKV matmul, Q/K per-head norm, RoPE, split-KV attention) for plain `@Parallel` reference implementations → still garbage.
- Swapped the RMS reductions alone → **correct output, exact logits match with the all-reference run** (`argmax='<think>'`, host-recomputed logits ≡ GPU logits).

**Fix.** New `reductionOneBlockWithLayerSingleGroup`: one workgroup strides over the input, reduces in local memory, lane 0 writes the final scale — race-free by construction, one kernel, no extra launch. The Qwen3 FP16 layer graphs use it on the NVIDIA scheduler path (grid `global == local == localSize`); the NON_NVIDIA path keeps the existing partials + finalize pair.

**Verified** (all coherent, greedy):
| model | before | after |
|---|---|---|
| Qwen3-1.7B FP16 | garbage (`\n\n\n…9…avier`) | `<think> Okay, the user is asking for the capital of France… Paris` (matches the validated batched-decode output) |
| Qwen3-1.7B + CUDA graphs | garbage | coherent |
| Qwen3-4B | coherent | coherent (unchanged) |
| Llama-3.2-1B / Mistral-7B | coherent | coherent (path untouched) |

Perf unchanged: Qwen3-1.7B ~51 tok/s, Llama-1B ~90 tok/s before/after.

**Note.** The same racy combine is still used by the Llama layer graphs and the logits layer (`TransformerComputeKernels.reductionOneBlockWithLayer`) — they currently win the race on this hardware, but they are latent instances of the same bug class; happy to extend the fix there if you want it in this PR.

### DIFF
--- a/src/main/java/org/beehive/gpullama3/tornadovm/kernels/TransformerComputeKernelsLayered.java
+++ b/src/main/java/org/beehive/gpullama3/tornadovm/kernels/TransformerComputeKernelsLayered.java
@@ -2489,6 +2489,44 @@ public class TransformerComputeKernelsLayered {
         return localSum[0];
     }
 
+    /**
+     * Race-free RMS-norm reduction in ONE workgroup: threads stride over the input, reduce in
+     * local memory, and lane 0 writes the final scale factor to {@code output[0]}.
+     *
+     * <p>The multi-workgroup variant ({@code reductionOneBlockWithLayer}) has workgroup 0
+     * combining the other workgroups' partial sums with <em>no inter-workgroup
+     * synchronization</em> — a data race. On the CUDA backend the race outcome is
+     * schedule/compilation dependent: e.g. Qwen3-1.7B reads stale partials on every layer
+     * (wrong normalization scale → garbage output) while Llama-1B and Qwen3-4B happen to win
+     * the race. The NON_NVIDIA scheduler avoids the race with a separate
+     * {@code reductionFinalNormalization} task; this kernel is the NVIDIA-path equivalent.
+     * Launch with exactly one workgroup: {@code global == local == localMemSize}.</p>
+     */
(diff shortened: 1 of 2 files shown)


mikepapadim:
@orionpapadakis is this an actual problem we solve?

orionpapadakis (approved):

PR #149, opened 2026-08-28
Refactor/framework abstractions

orionpapadakis:
Turns the TornadoVM-based LLM runner into a structured inference engine: a small public
façade, a backend-neutral program description, provider-discovered model families, and one
narrow SPI that every backend implements. TornadoVM remains the compiler and runtime — this
project builds no IR and generates no device code.

## What the engine now is

**A public façade, and nothing else reachable.** `LocalModels.load` returns a `LocalModel`
that is weights, configuration and identity: no forward, no sampler, no plan accessor.
Generation is a capability, so a model that only produces embeddings stops at `LocalModel`.
Thirteen types are stable; everything else reachable from the façade is marked experimental,
and a public type in neither set fails the build. No TornadoVM, GGUF or CLI type appears in
any public signature.

**A shared operation vocabulary.** A forward pass is an ordered list of components over
named weights — data, not execution. The CPU implementations and the generated kernels
compute the same operations rather than each carrying their own arithmetic, which is what
makes the CPU a usable numerical reference.

**Providers, not switches.** Recognition, loading, layer topology, the CPU forward pass,
task graphs, lowering and KV layout are each a `ServiceLoader` provider. Adding a family
means adding providers and service entries; adding a central switch fails an architecture
rule.

**One backend boundary.** TornadoVM types, task graphs, grid schedulers and execution plans
live inside the Tornado backend and nowhere else — both allowlists are empty. The backend
owns lowering, fusion and kernel selection, and selects kernels by declared device
capability rather than by testing a backend's name.

**Pooled KV and a shared device workspace.** Storage belongs to a cache manager and sessions
hold leases. The pool and block table are each one persistent device array walked in-kernel,
because captured-graph replay bakes addresses and re-pointing a buffer produces wrong output
rather than an error. A compiled program owns the workspace and is built once; an invocation
writes values into persistent control arrays and rebinds nothing, so several sessions share
one program and one device copy of the weights.

**An engine tier** above sessions for admission, batching and prefix reuse, which nothing
below it may depend on.

**No silent fallback.** An unsupported device selector, lowering combination or storage
request fails by name.

## JDK 21 and JDK 25

Each line publishes its own artifact against its own TornadoVM line — `-jdk21` against
`6.0.0-jdk21`, `-jdk25` against `6.0.0-jdk22plus` — with profiles pinned to exactly one JDK
each and `maven-enforcer` rejecting anything else at `validate`. `--enable-preview` is JDK 21
only, so the JDK 25 artifact carries no preview class file and is not pinned to one JDK
version.

Both launchers take their JVM flags from the SDK's own `tornado-argfile` instead of a
hand-maintained list. The JDK 25 line needs a patched JVMCI module and must not be given the
preview flag; neither was in the hardcoded list, which also passed `-XX:+EnableJVMCI`
unconditionally.

## Verification

Measured on an RTX 5090 Laptop against TornadoVM 6.0.0 built from the pinned tag:

| Configuration | Unit | Accelerator |
| --- | --- | --- |
| JDK 21 / CUDA | 531 pass | 87 tests, 0 failures, 6 skipped |
| JDK 25 / CUDA | 531 pass | 87 tests, 0 failures, 6 skipped |
| JDK 21 / OpenCL | 531 pass | 87 tests, 3 failures, 7 skipped |
| JDK 25 / OpenCL | 531 pass | 87 tests, 3 failures, 7 skipped |

CI compiles and unit-tests both JDK lines on both Linux backends and on the Metal SDK, and
asserts what each build produced: artifact suffix, class-file version, service-file count,
and a launcher command built from the SDK argfile with no hardcoded JVMCI flag. All twenty
standalone CUDA rows and all twenty OpenCL rows pass with a resolved backend, a real
execution path and the expected substring in the generated text.

Numerical claims are scored against the CPU, never GPU-against-GPU: two paths that are
equally wrong agree with each other and both pass.

### Known limitations, recorded rather than gated away

- **OpenCL, Llama F16, CPU parity.** About 21% of logit elements fall outside the tolerance,
  worst case 10.6×. Generated text is still correct and every other row passes, including
  Qwen3 F16, Qwen2.5 F16 and both Q8_0 rows. It is not the lowered path and not the
  single-pass RMS capability; CUDA and Metal are unaffected. The accelerator suite had never
  been run on this backend before, so this is newly measured rather than a regression.
- **The CPU-parity suite covers Llama, Qwen3 and Qwen2.5 only.** Granite, Phi-3, Mistral,
  DeepSeek-R1-Distill-Qwen and Devstral have fixtures, CPU equivalence tests and CI rows, but
  their GPU output is not gated against the CPU reference.
- **Batched prefill on Metal, and Q8_0 batched prefill on CUDA**, are TornadoVM toolchain
  gaps with named causes. Batched prefill stays default-off on every backend.
- **Kernel capture on Metal** produces no source, so compiled-program identity cannot be
  observed there.
- **Devstral** loads and generates correct text on Metal in `STANDARD`; the rest of its
  acceptance is unrun and it is not claimed as verified.
- **Memory preflight on Metal** is capped at `CONSERVATIVE`.

Metal evidence is CI and Mac-session evidence. Nothing here claims a Metal run performed
from Linux.

## Documentation and integrations

`docs/architecture` is eight documents describing the engine as it is — layers, execution,
the stable surface, ownership and concurrency, families and backends, verification, and
development. The design history is in git.

Both downstream integrations build against the frozen surface, importing only `api/**` plus
the two types marked experimental, and their CI job runs against the branch that tracks it.


orionpapadakis:
## performance preview

A snapshot of where this branch stands on throughput, taken because the refactor has now changed two things that could plausibly move it: the lowered execution path became the default for one combination (T13.6 / ADR-021), and batched prefill was corrected (DEF-01).

**This is a preview, not a gate.** One laptop, one model size, a short prompt. It is not the CI baseline in `docs/perf-history.jsonl`, which was recorded on different hardware and is not comparable to these numbers.

### Environment

| | |
| --- | --- |
| GPU | NVIDIA GeForce RTX 5090 Laptop |
| Backend | CUDA |
| TornadoVM | 5.2.1-jdk21-dev |
| Model | `Llama-3.2-1B-Instruct`, F16 and Q8_0 |
| Generation | 256 tokens, fixed seed, `--gpu-memory 14GB` |
| Method | 2 warm-up runs discarded, 5 measured, **median** reported, spread = (max−min)/median |

Every run's execution path was read back from the run metrics (`execution_path`), not inferred from the flag it was launched with — see the note at the bottom on why that distinction is load-bearing.

### Decode throughput (`eval_rate`)

| Configuration | median tok/s | spread | path |
| --- | ---: | ---: | --- |
| F16 single-token, `auto` | **142.39** | 0.9% | lowered |
| F16 single-token, `off` | 142.21 | 2.8% | legacy |
| F16 sequential prefill | 140.78 | 1.4% | legacy |
| F16 batched prefill (8) | 140.12 | 0.7% | legacy |
| Q8_0 single-token, `auto` | **151.00** | 1.0% | legacy |
| Q8_0 sequential prefill | 150.51 | 0.5% | legacy |
| Q8_0 batched prefill (8) | 149.20 | 1.8% | legacy |

Decode rate is essentially flat across modes, which is expected — all three decode one token at a time through the same kernels once the prompt is in the KV cache.

### Prompt processing throughput (`prompt_eval_rate`)

This is where the modes actually differ:

| Model | single-token | sequential prefill | batched prefill (8) |
| --- | ---: | ---: | ---: |
| F16 | 193.57 | 234.63 | **332.83**  (+72% vs single-token) |
| Q8_0 | 203.23 | 250.38 | **263.94**  (+30% vs single-token) |

**Caveat that matters:** the benchmark prompt is short (~34 tokens), so batched prefill gets between three and five chunks at batch 8. Its advantage should grow with prompt length and shrink toward the sequential number on very short prompts. Treat these as a lower bound on the batching win, not a headline.

### The lowering cutover is performance-neutral

`llama.lowering` is now tri-state (`auto` / `on` / `off`), and under `auto` exactly one combination — Llama / F16 / STANDARD — takes the lowered path. That qualification required a paired A/B, run as **one build differentiated only by the property**, interleaved so thermal drift hits both legs equally:

| Run | lowered | legacy | delta | tolerance | verdict |
| --- | ---: | ---: | ---: | ---: | --- |
| Cutover gate (5 pairs) | 144.37 | 144.62 | **−0.2%** | 3.0% | PASS |
| Re-check after DEF-01 (3 pairs) | 146.74 | 146.53 | **+0.1%** | 3.0% | PASS |

The two runs differ from each other by ~2 tok/s in absolute terms while agreeing on the delta, which is a fair illustration of why paired measurement is used here: an absolute number compared against one taken an hour earlier is not a measurement on this machine.

Every other combination stays on the legacy path and is unaffected.

### DEF-01

Batched prefill's logits used to leave the FP16 parity bounds by **27.80105745315552**. That is fixed: it now reproduces the single-token path **bit-for-bit** (worst excess `0.0`, original bounds unchanged) at batch 2, 8 and 64.

It was not a kernel defect. The batched branch prefilled the whole prompt including its last token and then decoded starting from that same token again, so the model saw `prompt + duplicate(last token)`. No kernel, task graph, workgroup, fusion or accumulation order was touched — the fix is host-side sequence construction. Full write-up in `docs/architecture/review/DEF-01-batched-prefill-diagnosis.md`.

**Batched prefill remains off by default.** The numbers above show it is worth having, but enabling it is a separate decision that needs its own paired evidence, and correctness alone does not make that decision.

### On reading the path back rather than trusting the flag

Earlier in this branch an entire accelerator gate passed green while the lowering flag never reached the JVM — a `-DargLine` override had replaced the profile's arguments, so a paired A/B was measuring legacy against legacy. Every measurement above reports which path actually ran, and the nine parity tests that used to establish their "legacy" reference by *clearing* the property now pin it explicitly, because under `auto` clearing it no longer means legacy.

### What is not claimed here

- No multi-GPU, multi-backend or larger-model coverage — this is 1B on one device.
- Not a regression gate; `scripts/perf-gate-run.sh` and `scripts/perf-ab-run.sh` remain the gates.
- Absolute numbers drift several percent across a long session on this machine. The deltas are the trustworthy part; the absolutes are indicative.

orionpapadakis:
## correction to the performance preview — DEF-01's parity claim

The [performance preview](https://github.com/beehive-lab/GPULlama3.java/pull/149#issuecomment-5513744329) said batched prefill "now reproduces the single-token path **bit-for-bit**". **That was overstated.** The corrected measurement, from the closeout regression:

| batch | rows | rows out of bounds | max abs diff | margin to bound | bit-identical floats |
| ---: | ---: | ---: | ---: | ---: | ---: |
| 2 | 64 | **0** | 0.006325 | 0.0101 | 0.010% |
| 8 | 64 | **0** | 0.006325 | 0.0101 | 0.010% |
| 64 | 64 | **0** | 0.006325 | 0.0101 | 0.010% |

Batched prefill agrees with single-token **comfortably inside the FP16 parity bounds** — max absolute difference 0.006325 against `atol = 1.5e-2` — but the two paths are *not* bit-identical, and about 0.010% of individual logits match exactly. Residual FP16 rounding differences between different kernels are expected; the previous claim implied there were none.

**How the error happened**, since it is the interesting part: the diagnostic probe tracked "worst excess", defined as `|diff| − (atol + rtol·|ref|)`, in an accumulator initialised to `0`. Every excess was negative — that is what passing means — so the accumulator never moved, and the probe printed `0.0`. I read "no violation" as "no difference". The closeout regression measures the absolute difference directly, which is what surfaced it.

**Nothing about the verdict changes.** DEF-01 is fixed, the parity bounds were never touched, 0 of 64 rows are out of bounds at every batch size, and the gate asserting it is live in `ThreePhaseModeAccelTest`. The defect and its host-staging root cause are exactly as described. Only the strength of the agreement claim was wrong, and the repository records are corrected.

Batched prefill remains **default-off**; enabling it is still a separate decision needing its own performance evidence.

orionpapadakis:
## Performance matrix on this branch

Every supported model × quantization × execution mode × backend, measured on one machine at
`4f57347f`. 168 configurations; **88 produced a number**, 76 refused with a declared
`UnsupportedOperationException`, and 4 hit a real backend compiler failure (below).

**Machine** — Intel Core Ultra 9 275HX, NVIDIA RTX 5090 Laptop (24 GB), JDK 21.0.2, TornadoVM
6.0.0 (`jdk21` line), separate CUDA and OpenCL SDK builds.

**Method** — one measurement method for every backend, so the columns are comparable: the
ordinary generation path with `-Dllama.bench.ignoreEos=true`, numbers read from the metrics
JSON (`prompt_eval_rate` / `eval_rate`), never from wall-clock. The same ~46-token prompt, a
256-token total budget (~211 decoded tokens), fixed seed, one discarded warm-up, median of 3
measured runs. `--bench` is deliberately not used — it only drives the Tornado forward path and
cannot measure CPU, so it cannot produce a comparable CPU column.

Each run's `execution_path` was recorded rather than assumed from the flag. Every CPU row
reports no accelerator plan, which is what distinguishes a genuine CPU run from a silent
fallback. On the accelerated backends only Llama-3.2-1B F16 takes the lowered program path;
every other configuration still runs the legacy path.

#### Decode throughput, tok/s (standard mode)

| model | quant | CPU | OpenCL | CUDA | CUDA + graphs |
| --- | --- | ---: | ---: | ---: | ---: |
| Llama-3.2-1B-Instruct | F16 | 24 | 97 | 141 | 152 |
| Llama-3.2-1B-Instruct | Q8_0 | 36 | 117 | 150 | 161 |
| Qwen3-0.6B | F16 | 42 | 90 | 181 | 222 |
| Qwen3-0.6B | Q8_0 | 57 | 98 | 201 | 246 |
| Qwen2.5-1.5B-Instruct | F16 | 19 | 55 | 51 | 54 |
| Qwen2.5-1.5B-Instruct | Q8_0 | 29 | 63 | 53 | 56 |
| Granite-4.0-1B | F16 | 17 | 44 | 49 | 53 |
| Granite-4.0-1B | Q8_0 | 27 | 52 | 50 | 54 |
| Granite-3.2-2B-Instruct | F16 | 12 | 45 | 69 | 78 |
| Granite-3.2-2B-Instruct | Q8_0 | 19 | 56 | 73 | 82 |
| Phi-3-mini-4k-instruct | F16 | 8 | 45 | 55 | 60 |
| Phi-3-mini-4k-instruct | Q8_0 | 13 | 55 | 62 | 66 |
| Mistral-7B-Instruct-v0.3 | F16 | 4 | 13 | 18 | 19 |
| Mistral-7B-Instruct-v0.3 | Q8_0 | 7 | 14 | 19 | 20 |

#### Prefill throughput, tok/s (standard mode)

| model | quant | CPU | OpenCL | CUDA | CUDA + graphs |
| --- | --- | ---: | ---: | ---: | ---: |
| Llama-3.2-1B-Instruct | F16 | 23 | 109 | 192 | 234 |
| Llama-3.2-1B-Instruct | Q8_0 | 34 | 134 | 205 | 247 |
| Qwen3-0.6B | F16 | 37 | 88 | 177 | 237 |
| Qwen3-0.6B | Q8_0 | 54 | 94 | 188 | 261 |
| Qwen2.5-1.5B-Instruct | F16 | 18 | 71 | 111 | 130 |
| Qwen2.5-1.5B-Instruct | Q8_0 | 26 | 83 | 120 | 140 |
| Granite-4.0-1B | F16 | 15 | 54 | 101 | 121 |
| Granite-4.0-1B | Q8_0 | 24 | 67 | 106 | 125 |
| Granite-3.2-2B-Instruct | F16 | 11 | 48 | 88 | 110 |
| Granite-3.2-2B-Instruct | Q8_0 | 17 | 62 | 95 | 117 |
| Phi-3-mini-4k-instruct | F16 | 7 | 47 | 72 | 78 |
| Phi-3-mini-4k-instruct | Q8_0 | 12 | 58 | 81 | 85 |
| Mistral-7B-Instruct-v0.3 | F16 | 3 | 24 | 38 | 41 |
| Mistral-7B-Instruct-v0.3 | Q8_0 | 6 | 32 | 43 | 45 |

#### Prefill throughput by execution mode, tok/s

Only families with a prefill/decode implementation are listed; the rest are in the coverage table below.

| backend | model | quant | standard | prefill-decode | batch-prefill-decode |
| --- | --- | --- | ---: | ---: | ---: |
| cpu | Llama-3.2-1B-Instruct | F16 | 23 | 28 | 42 |
| cpu | Llama-3.2-1B-Instruct | Q8_0 | 34 | 40 | 50 |
| cpu | Granite-4.0-1B | F16 | 15 | 16 | 16 |
| cpu | Granite-4.0-1B | Q8_0 | 24 | 24 | 24 |
| cpu | Granite-3.2-2B-Instruct | F16 | 11 | 11 | 11 |
| cpu | Granite-3.2-2B-Instruct | Q8_0 | 17 | 16 | 17 |
| opencl | Llama-3.2-1B-Instruct | F16 | 109 | 123 | 328 |
| opencl | Llama-3.2-1B-Instruct | Q8_0 | 134 | 154 | 286 |
| opencl | Qwen3-0.6B | F16 | 88 | 100 | 692 |
| opencl | Qwen3-0.6B | Q8_0 | 94 | 107 | 463 |
| cuda | Llama-3.2-1B-Instruct | F16 | 192 | 234 | 1115 |
| cuda | Llama-3.2-1B-Instruct | Q8_0 | 205 | 249 | **fail** |
| cuda | Qwen3-0.6B | F16 | 177 | 223 | 723 |
| cuda | Qwen3-0.6B | Q8_0 | 188 | 247 | **fail** |
| cuda-graphs | Llama-3.2-1B-Instruct | F16 | 234 | 288 | 1186 |
| cuda-graphs | Llama-3.2-1B-Instruct | Q8_0 | 247 | 309 | **fail** |
| cuda-graphs | Qwen3-0.6B | F16 | 237 | 321 | 1356 |
| cuda-graphs | Qwen3-0.6B | Q8_0 | 261 | 378 | **fail** |


#### Execution-mode coverage

Prefill/decode is implemented per family, not universally, and refuses by name rather than
falling back silently. This is existing coverage, unchanged by this branch — the matrix just
makes it explicit.

| family | CPU prefill/decode | GPU prefill/decode |
| --- | --- | --- |
| Llama | yes | yes |
| Granite | yes | no |
| Qwen3 | no | yes |
| Qwen2 | no | no |
| Phi3 | no | no |
| Mistral | no | no |

#### The one real failure

`batch-prefill-decode` + `Q8_0` on the **PTX backend only** aborts at compile time:

```
uk.ac.manchester.tornado.api.exceptions.TornadoInternalError:
unimplemented: address origin unimplemented: tornado.graal.compiler.nodes.calc.MulNode
```

Four configurations (Llama-3.2-1B and Qwen3-0.6B, Q8_0, with and without CUDA graphs). The same
model, mode and quantization compiles and runs on OpenCL, and F16 compiles on PTX, so this is
the PTX code generator failing to fold a multiply appearing in an address computation that only
the Q8_0 batched kernel produces. Batched prefill remains default-off, so nothing ships broken.

#### Two numbers that look wrong and are not yet explained

- Granite-4.0-1B decodes at 49 tok/s on CUDA while the larger Granite-3.2-2B reaches 69.
- Qwen2.5-1.5B decodes no faster on CUDA (51) than on OpenCL (55).

Both are reproducible across all three measured runs. Neither is a correctness problem — both
families hold CPU parity — but both suggest kernel selection worth profiling separately.

#### What the numbers say

- CUDA graphs are a consistent win, never a regression across all 44 comparable pairs: +2.8% to
  +24.1% decode and +4.7% to +87.7% prefill.
- Batched prefill is the largest single effect where it is supported — 5.8× prefill on
  Llama-3.2-1B F16 on CUDA graphs (234 to 1186 tok/s), 5.7× on Qwen3-0.6B — at a small,
  consistent decode cost (~1%).
- Q8_0 beats F16 for decode everywhere, and for prefill everywhere except OpenCL batched
  prefill, where F16 is faster on both supported families (Qwen3-0.6B 692 vs 463 tok/s).

orionpapadakis:
### Follow-up: the PTX failure above is fixed

`2b462d27`. The four `batch-prefill-decode` + `Q8_0` configurations that aborted on CUDA now run.

**Cause.** The Q8_0 batch-prefill kernels take the tensor-core path only where `TENSOR_CORE_MMA`
is granted, which is CUDA alone — that is why OpenCL was unaffected and why F16 was fine, since
the F16 tile never converts anything. In `packQ8Halves` the dequantized weight goes through
`new HalfFloat(v).getHalfFloatValue()`: the half is read back as a short and packed into a
register pair rather than stored into a `HalfFloatArray`, the object survives escape analysis,
and the PTX backend cannot lower the resulting address.

Isolated by bisecting the kernel — forcing the non-MMA path made Q8_0 batch prefill compile and
run on CUDA, and removing only the `HalfFloat` allocation from the MMA kernel made it compile
with everything else intact.

**Fix.** `fp16BitsOf` computes the half's bit pattern arithmetically, no allocation. It has to
round exactly as the platform did — a helper that truncated would bias every Q8_0 weight in one
direction and cost the model its CPU parity without failing to build — so `Fp16BitsTest` holds it
against `Float.floatToFloat16` over two million values, subnormals and the tie-breaking and
saturation boundaries included. It is not a TornadoVM patch; nothing outside this repository
changed.

**Verification.** All 11 CPU-parity cases green on CUDA and on OpenCL after the change, plus
`BatchedPrefillPolicyAccelTest`. New numbers for the four rows, same method as above:

| backend | model | quant | prefill tok/s | decode tok/s |
| --- | --- | --- | ---: | ---: |
| cuda | Llama-3.2-1B-Instruct | Q8_0 | 482 | 146 |
| cuda | Qwen3-0.6B | Q8_0 | 442 | 196 |
| cuda + graphs | Llama-3.2-1B-Instruct | Q8_0 | 495 | 158 |
| cuda + graphs | Qwen3-0.6B | Q8_0 | 457 | 238 |

Batched Q8_0 prefill lands well below the F16 tensor-core figures (1186 tok/s on Llama-3.2-1B)
because the dequantization sits in the GEMM's inner K-loop. That is a throughput question, not a
correctness one, and it is not addressed here.

With this, every failure in the matrix above is a declared `UnsupportedOperationException` for a
family × mode that has no implementation. No configuration fails unexpectedly on any backend.

orionpapadakis:
## Batched prefill: GPULlama3.java vs llama.cpp

Llama and Qwen3 — the two families with a GPU batched-prefill implementation — across every size
and quantization in the local model set. 17 models, 140 measurements.

**Machine** — Intel Core Ultra 9 275HX, NVIDIA RTX 5090 Laptop (24 GB), JDK 21.0.2,
TornadoVM 6.0.0 (`jdk21`). GPULlama at `696ed152`; llama.cpp at `968c43891` (build 9702), CUDA.

**Method** — each engine runs its *own* benchmark harness (`llama-bench` / `LlamaBench`):
synthetic token stream, no sampling, no chat template, one warm-up discarded, mean ± stddev over
3 repetitions. Both process the prompt in **512-token batches** — llama.cpp's default micro-batch
is 512 and GPULlama is given `-b 512` — so prefill is measured on the same operation. Decode is
`tg128` taken from that same batched run, which means it is the batched-prefill mode's decode
path, not the standard one. GPULlama is given `--gpu-memory 22GB`; the launcher's 14GB default is
less than an 8B fp16 model's weights, and leaving it would have measured the cap rather than the
engine.

### Prefill (prompt processing), tok/s

pp512 in 512-token batches on both engines. Higher is better.

| model | llama.cpp CUDA | GPULlama CUDA | GPULlama CUDA+graphs | GPULlama OpenCL | best GPULlama vs llama.cpp |
| --- | ---: | ---: | ---: | ---: | ---: |
| Llama-3.2-1B-Instruct-F16 | 32083 | 7922 | 7830 | 439 | 0.25× |
| Llama-3.2-1B-Instruct-Q4_K_M | 33795 | 4717 | 4810 | 376 | 0.14× |
| Llama-3.2-1B-Instruct-Q5_K_M | 32820 | 4754 | 4801 | 376 | 0.15× |
| Llama-3.2-1B-Instruct-Q8_0 | 36339 | 4974 | 4807 | 372 | 0.14× |
| Llama-3.2-3B-Instruct-F16 | 11341 | 3211 | 3239 | 141 | 0.29× |
| Llama-3.2-3B-Instruct-Q8_0 | 13670 | 1985 | 1988 | 131 | 0.15× |
| meta-llama-3.1-8b-instruct.fp16 | 5711 | 1689 | 1691 | 55 | 0.30× |
| meta-llama-3.1-8b-instruct.Q8_0 | 6386 | 1010 | 999 | 38 | 0.16× |
| Qwen3-0.6B-f16 | 36454 | 6169 | 6331 | 776 | 0.17× |
| Qwen3-0.6B-Q8_0 | 39842 | 4532 | 4631 | 614 | 0.12× |
| Qwen3-1.7B-f16 | 19135 | 4836 | 5021 | 332 | 0.26× |
| Qwen3-1.7B-Q4_K_M | 20772 | 3212 | 3221 | 271 | 0.16× |
| Qwen3-1.7B-Q8_0 | 22047 | 3222 | 3356 | 268 | 0.15× |
| Qwen3-4B-f16 | 8878 | 1950 | 1957 | 107 | 0.22× |
| Qwen3-4B-Q8_0 | 10469 | 1326 | 1333 | 103 | 0.13× |
| Qwen3-8B-f16 | 5238 | 1563 | 1565 | 54 | 0.30× |
| Qwen3-8B-Q8_0 | 6265 | 968 | 969 | 38 | 0.15× |

### Decode (token generation), tok/s

tg128, single-token decode over the growing KV cache, measured in the same batched run — so this is the batched-prefill mode's decode path.

| model | llama.cpp CUDA | GPULlama CUDA | GPULlama CUDA+graphs | GPULlama OpenCL | best GPULlama vs llama.cpp |
| --- | ---: | ---: | ---: | ---: | ---: |
| Llama-3.2-1B-Instruct-F16 | 297 | 197 | 219 | 113 | 0.74× |
| Llama-3.2-1B-Instruct-Q4_K_M | 658 | 209 | 236 | 145 | 0.36× |
| Llama-3.2-1B-Instruct-Q5_K_M | 613 | 212 | 237 | 144 | 0.39× |
| Llama-3.2-1B-Instruct-Q8_0 | 502 | 211 | 235 | 144 | 0.47× |
| Llama-3.2-3B-Instruct-F16 | 120 | 73 | 79 | 52 | 0.66× |
| Llama-3.2-3B-Instruct-Q8_0 | 200 | 82 | 88 | 66 | 0.44× |
| meta-llama-3.1-8b-instruct.fp16 | 54 | 39 | 41 | 28 | 0.76× |
| meta-llama-3.1-8b-instruct.Q8_0 | 95 | 44 | 46 | 37 | 0.49× |
| Qwen3-0.6B-f16 | 458 | 230 | 285 | 98 | 0.62× |
| Qwen3-0.6B-Q8_0 | 667 | 268 | 337 | 107 | 0.51× |
| Qwen3-1.7B-f16 | 207 | 144 | 168 | 68 | 0.81× |
| Qwen3-1.7B-Q4_K_M | 446 | 158 | 182 | 84 | 0.41× |
| Qwen3-1.7B-Q8_0 | 332 | 158 | 180 | 83 | 0.54× |
| Qwen3-4B-f16 | 95 | 71 | 78 | 40 | 0.82× |
| Qwen3-4B-Q8_0 | 157 | 79 | 85 | 52 | 0.54× |
| Qwen3-8B-f16 | 53 | 43 | 45 | 27 | 0.85× |
| Qwen3-8B-Q8_0 | 93 | 49 | 52 | 36 | 0.56× |

Every configuration produced a number: 140 measurements over 17 models, no failures.

### The 8B fp16 models: a device-memory defect, now fixed

These two models originally reported `oom` on all three GPULlama backends. That was a real defect,
not a hardware limit, and the numbers above are from after the fix (`a75e74a6`).

In `BATCH_PREFILL_DECODE` the plan builds a batch-prefill graph *and* a decode graph per layer, and
both bound that layer's weights with `transferToDevice(FIRST_EXECUTION, ...)`. A weight array bound
in two task graphs of one execution plan gets a device buffer in each, so the pool had to hold the
whole model twice — Llama-3.2-3B-F16 peaked at 12463 MiB against 5531 MiB in standard mode, 2.06×
its 5.99 GiB of weights. 16 GB of fp16 weights therefore could not fit a 24 GB card in batched mode
while running fine in standard mode.

The symptom hid the cause: the same 96 MiB allocation was refused at batch 32, 64, 128 and 512
alike, because the duplication is per-layer weights and has nothing to do with the batch.

The decode graphs now consume the copy their layer's prefill graph already uploaded, the way they
already consume the KV cache and the block table. Llama-3.2-3B-F16 batched peaks at 4505 MiB, down
from 12463. All 11 CPU-parity cases stay green on CUDA and on OpenCL.

### What the numbers say

- **Prefill is where the gap is largest**: GPULlama reaches 0.12×–0.30× of llama.cpp. The ratio is
  consistently better for F16 (0.22×–0.30×) than for quantized weights (0.12×–0.16×), because
  GPULlama dequantizes inside the GEMM's inner K-loop while llama.cpp uses dedicated quantized
  kernels.
- **Decode is much closer**: 0.36×–0.85×. The best cases are F16 (Qwen3-8B 0.85×, Qwen3-4B 0.82×,
  Qwen3-1.7B 0.81×) — the gap narrows as the model grows and decode becomes bandwidth-bound.
- **CUDA graphs help decode and not prefill.** Decode gains 7%–26% on every model; prefill moves
  by less than 3% either way. Prefill is already one large batched submission, so there is little
  launch overhead left to remove.
- **K-quants buy GPULlama nothing.** llama.cpp gets its fastest decode from Q4_K_M
  (Llama-3.2-1B: 658 vs 502 tok/s for Q8_0). GPULlama is flat across Q4_K_M / Q5_K_M / Q8_0
  (209 / 212 / 211), consistent with the loader mapping K-quants onto Q8_0 activations — the
  smaller weights never turn into less work.
- **OpenCL is far behind CUDA on prefill specifically**: 7.4×–26.5× slower for prefill against
  1.2×–2.5× for decode. The batched-prefill path leans on tensor-core MMA kernels that only the PTX backend
  lowers, so OpenCL falls back to a general GEMM.

.github/actions/setup-tornadovm/action.yml
```diff
@@ -1,45 +1,76 @@
...
+    default: '21'
 
 # macOS ships bash 3.2 (Apple keeps it frozen for GPLv3 reasons); SDKMAN's own scripts use
 # bash 4+ syntax (${var^^}) and crash under it, so every step below uses Homebrew's bash on
 # macOS runners instead (requires `brew install bash` on the runner) and plain "bash" on Linux.
 # GitHub's action-manifest parser doesn't support YAML anchors, so the expression is repeated
-# per step rather than shared via &shell/*shell.
+# per step rather than shared via &shell/*shell. Composite actions have no top-level `env:`
+# either, so the pinned commit (see the "Resolve pinned TornadoVM commit" step) is threaded through
+# via $GITHUB_ENV/$GITHUB_OUTPUT rather than a shared constant.
```
orionpapadakis: comments

.github/actions/setup-tornadovm/action.yml
```diff
@@ -1,45 +1,76 @@
...
+    # consumeFromDevice(producer, ...) resolved against the previously executed graph rather than
+    # the named one and imported its buffer unconditionally, so the second decode pass imported a
+    # null buffer and the next ALLOC dereferenced it. Fixed upstream in 8603a9fa2d (PR #996) and
+    # released in 6.0.0. 6.0.0 also fixes the sketch-phase LambdaForm hidden-class reader.
+    #
+    # Note the tag naming change: 6.0.0's own tag carries no -jdkNN suffix (5.x had v5.2.0-jdk21 /
+    # v5.2.0-jdk25), because 6.0.0 collapsed the per-JDK lines into jdk21 + jdk22plus.
+    #
+    # Moving to a newer TornadoVM is a maintainer decision, not a CI tweak — see
+    # docs/architecture/metal/START-HERE-MACOS.md's decision boundaries.
```
orionpapadakis: comments

.github/actions/setup-tornadovm/action.yml
```diff
@@ -48,17 +79,47 @@ runs:
...
+    # pull_graal_jars.py skips per *filename*, so the foreign jars were never displaced, the
+    # correct 23.1.0 jars were downloaded alongside them, and both were assembled into
+    # share/java/graalJars. The runner's copy carries a Graal packaged as module `tornado.graal`
+    # (the 6.0.x/develop flavour left over from before the version pin), so the boot layer ended
+    # up with org.graalvm.graphio in BOTH jdk.internal.vm.compiler and tornado.graal:
+    #   java.lang.LayerInstantiationException: Package org.graalvm.graphio in both module
+    #   jdk.internal.vm.compiler and module tornado.graal
+    # TornadoVM's own scripts/build-release-sdks.py clears graalJars/ before every build for
+    # exactly this reason. Do the same, and take the jars only from the pinned release —
+    # no dependence on runner-global state, nothing to clean by hand.
```
orionpapadakis: comments

.github/actions/setup-tornadovm/action.yml
```diff
@@ -72,6 +133,43 @@ runs:
...
+    #   java.lang.LayerInstantiationException: Package org.graalvm.graphio in both module
+    #   jdk.internal.vm.compiler and module tornado.graal
+    #
+    # Either packaging is legitimate on its own — the 5.2.0 release line pulled Graal 23.1.0 whose
+    # compiler jar declares jdk.internal.vm.compiler, while 6.0.0 vendors its own tornado.graal
+    # module (tornado-graal-23.1.0.jar, alongside collections/truffle-compiler/word). What must
+    # never happen is BOTH in one dist, which is what seeding graalJars from the runner-global
+    # /opt/graalJars used to cause. So the invariant asserted here is that they do not mix, rather
+    # than which one is present — which is why the pin move from 5.2.0 to 6.0.0, flipping the dist
+    # from the upgrade-module packaging to the vendored one, needs no change here.
```
orionpapadakis: comments

.github/workflows/build-and-run.yml
```diff
@@ -598,6 +379,85 @@ jobs:
...
+  # eval_rate against the most recent gate-passing entry for the same tuple in
+  # docs/perf-history.jsonl, failing on a regression beyond tolerance.
+  #
+  # Deliberately NOT --append: publish-performance-history below is what records runs,
+  # and two writers would race for the same file and double-count the same run.
+  #
+  # Push to main only, for the same reason that job is: a PR from a fork cannot be
+  # trusted with the history, and a gate that runs on unpinned hardware compares against
+  # the wrong baseline. scripts/perf-gate-tolerances.json is where a machine's tolerance
+  # (or record-only status) is declared.
```
orionpapadakis: comments

.github/workflows/build-and-run.yml
```diff
@@ -624,14 +484,22 @@ jobs:
 
       - name: Append to performance history
         run: |
+          # Gate tuple fields (M1.7): a row missing any of these can never serve as a
```
orionpapadakis: comments

.github/standalone-expectations.tsv
orionpapadakis: is that file needed? if yes is it updated?

mikepapadim (approved):

src/main/java/org/beehive/gpullama3/backend/tornado/kernels/Phi3Kernels.java
```diff
@@ -185,6 +185,50 @@ public static void fusedRmsNormQKVMatmulDirect(
...
+        for (int j = localId; j < dim; j += 32) {
+            sum += wqkv.get(rowOffset + j).getFloat32() * (rmsWeights.get(j) * scale * x.get(j));
+        }
+
+        sum += context.simdShuffleDown(sum, 16);
+        sum += context.simdShuffleDown(sum, 8);
+        sum += context.simdShuffleDown(sum, 4);
+        sum += context.simdShuffleDown(sum, 2);
+        sum += context.simdShuffleDown(sum, 1);
+
```
mikepapadim, line 220: 🚀🚀🚀🚀

PR #152, opened 2026-09-15
Rename GPULlama3.java to jllm

mikepapadim:
Renames the project from **GPULlama3.java** to **jllm**, across all four surfaces: the Java package tree, the Maven coordinate, the GitHub repo slug, and the CLI/runtime contract.

The old name had drifted from what this is. It no longer serves only Llama 3 — Qwen2/3, Mistral, Devstral, Phi3, Granite and Gemma4 all run on it — and the `GPU`/`.java` decoration reads as a demo rather than a JVM inference engine.

**This is a clean break**: no relocation POM, no compatibility aliases, no deprecation window. Existing releases stay resolvable at `io.github.beehive-lab:gpu-llama3` forever; downstream consumers migrate on their own schedule.

## What changed

| | From | To |
|---|---|---|
| Java package | `org.beehive.gpullama3` | `org.beehive.jllm` |
| Maven artifactId | `gpu-llama3` | `jllm` |
| Maven groupId | `io.github.beehive-lab` | **unchanged** |
| Repo | `beehive-lab/GPULlama3.java` | `beehive-lab/jllm` |
| Launchers | `llama-tornado`, `llamaTornado` | `jllm`, `jllm4j` |
| System properties | `llama.*` (23 keys) | `jllm.*` |
| Env | `LLAMA_ROOT`, `GPULLAMA_TEST_MODELS` | `JLLM_ROOT`, `JLLM_TEST_MODELS` |
| Fixture cache | `~/.gpullama3/test-models` | `~/.jllm/test-models` |
| Classes | `LlamaApp`, `LlamaBench` | `JllmApp`, `JllmBench` |

**Not renamed** — the Llama *model family*: every `Llama*` architecture class, `ModelType.LLAMA_3`, `LLAMA_3_PATTERN`, the golden fixture dirs, and the GGUF metadata keys.

## Reviewing this

Nine commits, and they are meant to be read in order. **Please do not squash-merge** — commit 1 is a pure `git mv` with zero content bytes changed, which is what lets git record 100%-similarity renames so `git log --follow` and `git blame` survive. Squashing fuses it with commit 2 and rename detection then depends on a similarity heuristic that small files fail.

Commit 1 does not compile on its own; a package rename cannot be staged incrementally, since the directory and the `package` declaration must agree. `git bisect` will land there.

## The dangerous part: the `llama.` prefix collision

Three namespaces share the literal `llama.`, and `llama.attention.` straddles two of them:

- `llama.attention.head_count` — **GGUF metadata**, read out of the model file
- `llama.attention.splitKv` — **our system property**

A prefix substitution (`s/llama\./jllm./`) would have silently broken model loading for every Llama and Mistral GGUF, and it would not have shown up in CI: a corrupted key only fails when a real GGUF header is parsed, which is an accelerator test that *skips* without fixtures.

So the rename used a closed allowlist of the 23 project-owned keys, guarded four ways:

- **G1** both GGUF loaders still hold 12 `llama.*` literals each
- **G2** no GGUF key appears anywhere under a `jllm.` prefix
- **G3** no project system property survives outside those two loaders
- **G4** `LlamaModelLoader.java` and `MistralModelLoader.java` have **zero changed lines** — proof by construction that the keys were never edited, not merely restored

*(Reviewer note: scope G4 by path. The glob `*ModelLoader.java` also matches the base `ModelLoader.java`, whose `llama.providers` is a system property and is correctly renamed.)*

## Bugs fixed along the way

- **`java -jar` has never worked.** The shade plugin declared `<mainClass>io.github.beehive-lab.gpullama3.LlamaApp</mainClass>` — not a legal Java identifier (hyphen in `beehive-lab`), naming a package that never existed. Now `org.beehive.jllm.JllmApp`, and it runs.
- **`Automatic-Module-Name` is now pinned** to `org.beehive.jllm`. With no `module-info`, JPMS derives it from the jar filename, so the artifactId change would otherwise have silently moved it from `gpu.llama3` to `jllm`.
- `GPULLlama3TypeException` (note the triple L) was declared and referenced nowhere. Deleted.
- The usage banner still read `Usage: jbang Llama3.java [options]`, inherited from mukel/llama3.java.
- The CI service-file guard keyed on `^META-INF/services/org\.beehive\.`, which survived the rename by accident. Tightened so it cannot silently count nothing.

## Verification

- **535 unit tests**, zero failures, zero skips — on **both** JDK 21 and JDK 25 profiles
- **ArchUnit runs 21 rules and 15 self-test rules against a non-empty package.** A stale `ROOT_PACKAGE` would make every rule pass vacuously against zero imported classes; `DependencyRulesSelfTest` exists for exactly this
- `StableSurfaceTest`'s two reflective `Class.forName` lookups and the four source-path-literal tests all assert non-vacuously (checked assertion counts, not just exit codes)
- `make test-scripts`: 41 tests OK
- Jar: `Main-Class: org.beehive.jllm.JllmApp`, `Automatic-Module-Name: org.beehive.jllm`, exactly 7 service files, `java -jar … --help` works
- `./jllm --show-command` emits `-Djllm.*` and no `-Dllama.*`; resolves `target/jllm-1.0.0-jdk21.jar` and `org.beehive.jllm.JllmApp`
- `spotless:check` clean

The `jllm4j` name is load-bearing and was verified on JDK 25. It is an implicitly-declared class, so the filename must be a valid Java identifier and must not end in `.java`. `jllm.java` fails with `illegal character: '#'` (javac stops stripping the shebang for `*.java`); `jllm-java` fails with `bad file name`. `jllm4j` works, and `lookupClass().getName()` returns it, so `--help` prints the new name with no code change.

The last commit is `spotless:apply`. The profile uses `ratchetFrom origin/main`, so this PR — which touches every file — triggers the first full-tree format check the repo has ever had. Almost all of it is re-wrapping caused by the shorter package name; the one substantive change is an unused import dropped from `LlamaChatFormat.java`.

## Left alone deliberately

**History** — `CHANGELOG.md` entries for released versions, `docs/perf-history.jsonl` (1614 rows carrying the old workflow name, which is what actually produced them; the perf gate keys on a tuple that excludes `workflow`), and `perf-results/**`, whose baseline README holds a captured stack trace with old package frames. That is evidence, not a stale reference.

**Names we do not own** — `langchain4j-gpu-llama3`, `model-providers/gpu-llama3`, `quarkus.langchain4j.gpu-llama3.*` config keys, `GPULlama3ChatModel` and friends, the `gpullama3@beehive-lab` jbang alias, `beehivelab/gpullama3.java-*` Docker images, the 8 `*-gpullama3java` Hugging Face collections, and `beehive-lab/docker-gpullama3.java`.

## Before merging

**`QUARKUS_LANGCHAIN4J_REF` now points at `jllm/facade-1.0.0`, which does not exist yet.** That branch has to be created on the quarkus-langchain4j fork with the dependency artifactId changed to `jllm`, or the Quarkus leg is red.

Two more, both of which fail silently:

- **Re-point required status checks** right after renaming the repo. They are stored as job-name strings; a context recorded as `GPULlama3 Build & Run / build-linux` can never be satisfied again and blocks every future PR on a check that never reports.
- **`ln -s ~/.gpullama3 ~/.jllm` on every self-hosted runner** — symlink, not move, so pre-rename branches keep working. Without it the golden and accel tests skip with a named reason and the job still goes green.

Renaming the repo *before* merging is recommended: the 13 rewritten slug guards point at `beehive-lab/jllm`, so on the old slug every guarded job skips and the PR goes green having run almost nothing.

`POST_MOVE_ACTIONS.md` (added in this PR) is the full checklist for everything outside this repo — Maven Central, the two upstream integrations, Docker Hub, jbang, Hugging Face, CLA-assistant, DeepWiki — with a "do not do" list at the end.


kotselidis:
Why in the launcher we have jllm4j?

mikepapadim:
> Why in the launcher we have jllm4j?

because we kept the jllm for the python launcher and the jllm4j is the java launcher

kotselidis (approved):

mairooni (approved):

orionpapadakis (approved):
branch now ready to merge. has absorbed latest main without conflicts and renames to jllm everywhere

mikepapadim:
@orionpapadakis why ci does not run?

PR #164, opened 2026-09-17
Gemma 4 on CUDA

### LINKED ISSUE #161: Gemma 4 performance

mikepapadim:
Gemma 4 was verified on the CPU and on Apple-Silicon OpenCL in #120 and never on CUDA. It turns out
it did not work there. This fixes two independent defects, makes a Q4_0 file loadable at all, adds
the parity gate the family never had, and then optimizes decode against a pinned `llama.cpp`
reference.

Decode went from **0.104x to 0.789x** of that reference at `tg128`, each step admitted on its own
interleaved measurement.

## The two defects

Both were invisible without a CPU/GPU parity gate, and both produced fluent, wrong text — the
failure mode that exits 0 at a normal throughput.

**The RMS reduction was paired with the wrong grid.** All four Gemma 4 layer classes hardcoded the
multi-workgroup `reductionOneBlockWithLayer` at their twelve reduction sites and never called
`rmsReduceKernel()`. Llama and Qwen3 call it. The selector exists because the pairing is a contract,
stated in `AbstractLogitsTaskGraph`: that kernel is only safe when a `reductionFinalNormalization`
follows it, which the NVIDIA path skips — and the worker these classes *did* inherit launches a
single 256-thread workgroup. So it summed the first 256 squares of a 1536-wide row and added five
slots no workgroup had written. Visible in the logits: 252179 of 262144 sat at the 30.0 soft-cap
against none on the CPU, so the argmax was chosen among saturated values separated by 0.003.

This is also why the family passed on Apple Silicon: the NON_NVIDIA path adds the finalize and takes
the multi-workgroup grid, so both halves of the contract held there.

**`gatherPerLayerTokenEmbeddingRow` had no callers.** A private method in `Gemma4.java`, never
invoked, and nothing else writes `wrapPerLayerTokenEmbedRow`. On the accelerator that buffer held
zeros for the whole run, so every layer's per-layer-embedding contribution was `(projection + 0) /
sqrt(2)` — missing half its input and scaled as though it had both. The host path does the same
gather inline, which is why only the device was wrong.

Located by comparing the two paths quantity by quantity at position 0, where attention is a single
key and RoPE is identity: Q, K and V agreed to ~1e-7, attention weights were exactly 1.0, the
residual was still exact after the attention branch and after the FFN branch, then 3.1e-2 out after
the per-layer-embedding branch — and inside it, the model-projection half agreed to 7e-7 while the
gathered half was all zeros.

## Q4_0

`per_layer_token_embd` is Q5_K in a Gemma 4 Q4_0 file, and `LongIndexedTensor` read F32, F16, BF16
and Q8_0 only, so the file was refused before any weight reached a device — on the host path too.
The decode is `Q5_KFloatTensor`'s, the accepted host decoder; only the indexing widens, and a test
asserts bit identity against it over adversarial bytes.

The device still runs that file **materialized as Q8_0** — the provider admits F16 and Q8_0 — so this
is Q4_0 support in the sense that the file works, not in the sense that its blocks reach the device.
Residency is separate work. The conversion is scored rather than hidden: relative L2 0.0116 with
0 of 63 argmax disagreements, recorded as its own `Q4_0_MATERIALIZED` envelope, measured first and
written after, with a note that a retained path should be scored against `Q8_0`'s bounds instead.

## The optimizations

Each has its own commit with its own measurement. All were chosen from a per-task profile and, where
it mattered, from Nsight Compute — not from the generic attack order, which turned out to be wrong
for this model twice.

**1. One workgroup per attention head** (`86e977da`). The emitted CUDA showed
`for (i = blockIdx.x*blockDim.x + threadIdx.x; i < 8; )` on a grid of 512 threads: eight of them did
every position and every head dimension and 504 fell through. 52.1% of decode kernel time. The grid
already provided the lanes; this uses them. +651.7% on decode at depth, 5 interleaved pairs.

**2. Split the window across workgroups at depth** (`192b89a1`). Attention cost still grew with
context — 0.0164 ms/call at average depth ~12 against 0.2031 ms at ~551, while `ffn_gate_up` stayed
flat. Phase 1 slices the window; phase 2 is `combineSplitKVAttention` unchanged, which is already
generic in `headSize` and `nSplits`. +74.2%.

**3. A wider launch shape for narrow projections** (`c431b065`). The plan said packed-integer dot
products next, on the Qwen3.8 precedent. Measurement said otherwise: `ffn_gate_up` and `vocab_proj`
already run at 75–79% of DRAM peak, so on Q8_0 weights there are no bytes for dp4a to save —
Qwen3.8's win came from Q4_0 weights being half the bytes. ncu found `matrixVectorGenericQ8Byte` at
block 32 / grid 1536 reaching 27.4% of peak against `fusedGateUpGeGLUQ8` at block 32 / grid 6144
reaching 74.9%, running the *identical* inner routine. The difference was rows. That routine already
reduces over an arbitrary block, so no kernel changed — only the launch shape. 27.4% → 62.4%
measured, +5.4% end to end.

## Measured

RTX 4090, TornadoVM 6.1.0 (tag v6.1.0), JDK 21, CUDA 12.6, driver 610.57.04.
Reference is a pinned `llama.cpp` at `f172be756`, `-DGGML_CUDA=ON`, same session, same flags,
re-measured at each milestone and reproducible across three runs (236.75 / 237.18 / 236.70 at tg128).

| workload | before | after | +`--cuda-graphs` | llama.cpp | ratio |
| --- | ---: | ---: | ---: | ---: | ---: |
| tg128 | 24.58 | 140.08 | 186.81 | 236.70 | **0.789** |
| tg128 @ d2048 | — | 108.69 | 135.64 | 231.69 | 0.585 |
| pp512 | 6.94 | 131.92 | 175.43 | 23053.07 | 0.0076 |

`--cuda-graphs` is opt-in and off by default, so the default configuration is the third column, not
the fourth. Measured separately at +24.9% here and +2.6% on Llama-3.2-3B, both stable; its default
is read in the *shared* single-token plan and is deliberately **not** flipped in this PR — two
families is not evidence for the six others that run that plan on CUDA.

`llama.cpp`'s `pp512` carries ~14% spread, which fails the repo's own 10% stability gate, so that
column is an order of magnitude rather than a ratio I would defend.

Prefill moves only because this family declares `STANDARD` only, so prefill *is* the decode path run
once per prompt token. Closing it needs `BATCH_PREFILL_DECODE`, which does not exist for Gemma 4;
for calibration, #150 gave Qwen3.8 batched prefill, tensor cores and tiling and reported ~0.20x at
pp381.

## Correctness

`Gemma4CpuGpuParityAccelTest` is new — the family had no CPU/GPU gate, and a GPU-versus-GPU
comparison cannot see a defect that moves the whole GPU. Three fixtures of one checkpoint, pinned by
SHA-256. The Q4_0 case is a separate class because surefire forks per class and this family's three
fixtures cannot be loaded in one JVM on a 24 GB card: freed device memory returns to TornadoVM's
buffer provider, not the driver.

Every optimization is an arithmetic-order change and every one was scored with **bounds unchanged**:

| | Q8_0 | BF16 |
| --- | --- | --- |
| worst relative L2 | 2.80e-06 (bound 1e-4) | 7.41e-05 (bound 2e-3) |
| cosine | 1.00000000 | 1.00000000 |
| argmax disagreements | 0 / 63 | 0 / 63 |

Plus greedy generation token-identical to the recorded baseline on three prompts, and a 691-word
generation running past position 512 — so `windowStart > 0` is exercised rather than assumed — that
is byte-identical across the attention rewrites. 606 Class A tests pass, including the 21
dependency rules.

Both attention paths are reachable and both were confirmed from the emitted code rather than
inferred: at context 30, 35 `attentionWithSlidingWindowParallel`; at context 1024, 35
`attentionWithSlidingWindowSplit` and 35 `combineSplitKVAttention`. The split threshold is 512 so
that `GoldenCapture`'s 512 parity gate scores the path the benchmarks run — a higher one would have
left it validating the single-pass kernel, which is the Qwen lesson about tests checking the scalar
path while benchmarks used MMA.

## Not done, deliberately

- Q4_0 device residency. `ffn_gate_up` and `vocab_proj` are already near the DRAM roofline, so the
  remaining decode headroom on Q8_0 is small; fewer bytes per weight is the next real lever and it
  only pays on the Q4_0 fixture.
- The FP16 layer class does not get the wider projection launch shape. It reaches those projections
  through a different inner routine, and a launch shape that pays for one kernel is not evidence for
  another. Unmeasured, so unchanged.
- CI rows, and the `models-and-backends.md` / `verification.md` updates.
- `keyValueLayerCount()` is not overridden, so the memory preflight still predicts the KV cache over
  all 35 layers where only 15 own one. Conservative, so it refuses rather than under-predicts.

## Reviewer notes

`make lint` cannot run here: Spotless's JGit does not resolve the `.git` file a worktree uses, so
formatting is unverified by machine on this branch — worth a check in CI.

---

# Part two: prefill

The decode work above left prompt processing untouched, and it was the larger gap by two orders of
magnitude: this family's prefill was its decode loop run once per prompt token, so every weight was
read once per token — 512 sweeps of the model to ingest 512 tokens. That is not a tuning problem,
and none of the techniques in #163 apply to it, because every one of them is an optimization *inside*
an already-batched tensor-core prefill. Gemma 4 had no batched path at all.

**Prefill went from 0.008x to 0.33x of the same reference at `pp512`** — 39.5x on Q8_0 and 30.3x on
Q4_0 against this family's own single-token path — and the batched path now serves both
representations.

## Measured

Idle GPU, reference re-measured in the same session, `--cuda-graphs` on both sides.
RTX 4090 · TornadoVM 6.1.0-cuda (`3c6c5f12c`) · JDK 21.0.2 · driver 610.57.04 · CUDA 12.6 ·
`llama-bench` f172be756.

| Q8_0 | single-token | batched (b=512) | llama.cpp | batched ratio |
|---|---:|---:|---:|---:|
| pp512 | 196.53 | **7764.35** | 23752.94 ± 10.7% | **0.327** |
| pp512 @ d2048 | 184.91 | **3466.59** | 21724.37 ± 6.7% | 0.160 |
| tg128 | 200.42 | 197.40 | 236.77 | 0.834 |
| tg128 @ d2048 | 186.46 | 182.41 | 231.65 | 0.787 |

| Q4_0 | single-token | batched (b=512) | llama.cpp | batched ratio |
|---|---:|---:|---:|---:|
| pp512 | 240.47 | **7289.80** | 24934.03 ± 7.5% | **0.292** |
| pp512 @ d2048 | 223.67 | **3433.84** | 22494.04 ± 6.5% | 0.153 |
| tg128 | 245.86 | 240.59 | 324.19 | 0.742 |
| tg128 @ d2048 | 226.82 | 220.02 | 314.99 | 0.698 |

The `pp512` reference spreads exceed the 10% gate on Q8_0, so that ratio carries the reference's
uncertainty; the `d2048` references pass it.

## The steps

| commit | change | pp512 |
|---|---|---:|
| — | the decode loop, once per prompt token | 145 |
| `3ef3741c` | a batched prefill plan at all | 2767 |
| `a8a3706b` | attention keys staged through a transposed shared tile | 3153 |
| `fb2b002f` | depth split across blocks for the two narrow projections | 3797 |
| `bd490dbb` | gate/up decoded to FP16 once, then one FP16 GEMM | 4575 |
| `e4a59399` | query/key/value decoded the same way | 4771 |
| `201c9ec1` | an FP16 split-K, so the last two projections get both fixes | 5857 |
| `aaf67b7e` | attention key tile halved — occupancy, not coalescing | 6469 |
| `79a57fdf` | per-layer embedding rows gathered across tokens | 7510 |
| `d0252706` | the same path for Q4_0, with 18-byte and 20-byte decoders | — |

Only the first is a port of #163's structure. The rest came from measuring this family, and twice the
measurement contradicted what the ladder predicted:

- **`aaf67b7e`** was going to be #163's warp-per-dot-product attention rewrite. Nsight said the kernel
  was neither bandwidth-bound (DRAM 0.70%, L2 hit rate 99.55%) nor compute-bound, but occupancy-bound
  at 41.7% *"limited by the required amount of shared memory"* — by the key tile added two commits
  earlier. Halving it took theoretical occupancy to 75% and the kernel from 776.6 to 551.3 µs. One
  constant, +9.5%, bit-identical.
- **`79a57fdf`** was not a kernel at all. The profile said 230 ms of kernel time per four chunks; the
  wall clock said 98 ms per chunk. The difference was the host-side per-layer embedding gather —
  31.77 ms per chunk on Q4_0, counted by no profiler in this repository.

Reading the emitted CUDA was worth more than any profile. `gemmMMAGateUpQ8` is **8,666 lines carrying
sixteen `mma.sync`**, against 1,179 for the FP16 `gemmMMA`: the Q8_0 GEMMs convert float to half in
software, because TornadoVM reaches the hardware conversion only through a store to a half array.
Filed upstream as beehive-lab/TornadoVM#1096 and #1097 with minimal reproducers; #1097 also confirms
#1089 is still present at 6.1.0.

## What is Gemma 4's and what is reused

The GEMMs are this repository's own — `gemmMMA`, `gemmMMAQ8`, `gemmMMAQKVQ8`, `gemmMMAGateUpQ8` are
written in M, N and K and know nothing about any architecture — and so are the RMS reductions and the
FP32-to-FP16 cast. `Gemma4BatchPrefillKernels` holds only what has no counterpart elsewhere: sandwich
norms that normalize a branch output and add it onto the residual, a per-head norm on the query, key
*and* value, a GeGLU where every other family has SwiGLU, an attention scale of exactly one, and a
sliding window over a **flat** cache addressed by a per-layer base offset, where every batched
flash-attention kernel in the tree is paged.

Two head widths, two feed-forward widths, and twenty of thirty-five layers that project no key and no
value. Buffers are allocated at the widest and addressed with the layer's own stride; worker grids are
built per layer, because a grid keyed on a task name is wrong the moment one name maps to two shapes.
The Q4_0 file is mixed — `ffn_down` is Q4_1 on blocks 0-3 and Q4_0 on the other thirty-one — so the
decoder comes from each tensor's own representation, not the model's.

**No new dispatch classes.** An earlier shape of this added four and `DispatchLedgerTest` refused it;
the ledger stays at `{fp16=30, q8_0=32}`.

## Validation

- **Teacher-forced NLL against the host CPU**, which shares no kernel with any of this. Prefix 320 at
  chunk width 256, so it crosses a boundary. Pooled: Q8_0 host 6.58276 → batched 6.58219 (−0.00057
  nats/token); Q4_0 6.31673 → 6.30038 (−0.01635). Favourable on both, per-passage signs mixed. Two
  builds on fixed inputs, not a quality benchmark.
- **Needle-in-haystack across chunk boundaries.** A fact at ~token 250 of a ~1500-token prompt,
  recalled at the end, correct on both fixtures at widths 256 and 512 and on the single-token path.
  The sliding-window layers cannot see that far, so this is the one-in-five global layers, and it
  shows the KV cache and the SWA/global split both survive chunking.
- Short prompts (1-4 tokens, a chunk that is almost all padding rows) answer correctly on both.
- Both parity gates unchanged, bounds not touched; 606 Class A tests pass.

Two latent defects were found by comparing activations rather than by any benchmark, both producing
fluent, wrong output: decode layer 0 consuming its activation by the unnamed form — an activation of
exactly zero out of every decode layer — and the family ignoring the phase strategy, so the batched
plan was built and never driven.

## Corrections to earlier numbers in this PR

1. `3ef3741c`'s message says decode in the batched plan costs 147.08 against the single-token path's
   201.24. That compared a batched number measured *without* `--cuda-graphs` to a single-token number
   measured *with* it. Properly measured, graphs on both sides: **197.40 against 200.42 on Q8_0
   (−1.5%) and 240.59 against 245.86 on Q4_0 (−2.1%)**. The batched plan does not meaningfully cost
   decode.
2. **The decode campaign's figures were measured with `--cuda-graphs` and the prefill campaign's
   without.** Same commit, same GPU: `tg128` is 133.75 without and 176.39 with. HEAD reproduces the
   decode campaign's 201.24 / 185.58 exactly (200.42 / 186.46) under that flag, and `717881b3` — the
   commit before any prefill work — measures 147.96 against HEAD's 147.97 without it. Decode has not
   regressed; the two halves of this PR were reported in different configurations.

## Not done, in part two

- **Prefill is 0.29-0.33 of the reference, not the 0.6 that was the target**, and 0.15-0.16 at depth
  2048, because attention grows with depth and the kernel is still three passes over a global score
  scratch. #163's tensor-core attention is the next step and was not reached.
- FP16/BF16 has no batched components; the path is Q8_0/Q4_0/Q4_1, refused by name otherwise.
- The batched plan is opt-in (`--with-prefill-decode --batch-prefill-size N`), so nothing regresses
  unless it is asked for, and nothing improves either.

https://claude.ai/code/session_014xVfr1SjQPTruQos3HArW9


orionpapadakis (approved):

mikepapadim:
## Prefill: a batched plan for Gemma 4 (`3ef3741c`)

Taking the structural idea from #163 rather than its kernels. Every optimization retained there — paired 16-bit loads, the scored attention scratch, block lookahead, dequantize-then-GEMM, the tiled B-stage, tensor-core attention — is a change *inside* an already-batched tensor-core prefill. Gemma 4 had no batched path at all: its prompt processing was its decode loop run once per prompt token, so every weight was read once per token. That is the whole gap, and none of #163's techniques can be applied until it is closed.

### Measured (RTX 4090, TornadoVM 6.1.0-cuda, JDK 21, KV FP32, `gemma-4-E2B-it-Q8_0.gguf`)

| test | before | after (b=512) | factor | ratio to llama.cpp before → after |
|---|---:|---:|---:|---|
| pp512 | 145.03 | **2766.95 ± 109.39** | **19.1×** | 0.0061 → **0.116** |
| pp512 @ d2048 | 139.49 | **1377.56 ± 1.05** | **9.9×** | 0.0064 → **0.063** |

Chunk width is the whole story: b=128 → 1188.22, b=256 → 1938.46, b=512 → 2766.95.

Reference `llama-bench` f172be756, same session, same file, `-r 5`: pp512 **23786.58 ± 2693.67** — **±11.3%, which fails the 10% spread gate, so the d0 ratio carries that uncertainty**; pp512@d2048 21832.17 ± 1500.75 (±6.9%) passes. The jllm figures are 5 repetitions each, spread 4.0% and 0.1%.

For scale: the same batched machinery on Llama-3.2-3B-Q8_0 measured here gives 85.76 → 3102.89 t/s, i.e. 0.120 of its own `llama-bench` reference. Gemma 4 at 0.116 is now level with what this repository's batched path reaches elsewhere, which is where #163's ladder becomes applicable.

### What is reused and what is this family's

The GEMMs are the repository's own — `gemmMMAQ8`, `gemmMMAQKVQ8`, `gemmMMAGateUpQ8`, `gemmMMA` are written in M, N and K and know nothing about any architecture — and so are the RMS reductions and the FP32→FP16 cast. `Gemma4BatchPrefillKernels` holds only what has no counterpart elsewhere: sandwich norms that normalize a branch output and add it onto the residual, a per-head norm on the query, key *and* value, a GeGLU where every other family has SwiGLU, an attention scale of exactly one, a sliding window over a **flat** cache addressed by a per-layer base offset (every batched flash-attention kernel in the tree is paged), and the per-layer-embedding block.

Two head widths (256 sliding / 512 global), two feed-forward widths, and twenty of thirty-five layers that project no key and no value. Buffers are allocated at the widest and addressed with the layer's own stride; worker grids are built per layer, because a grid keyed on a task name is wrong the moment one name maps to two shapes.

### Correctness, before any timing was taken

Chunk-wide activations read back and compared against the single-token path position by position — rows 0–12 match to four decimals (row 0 rms 0.579464 vs 0.579381) — and the decode that follows matches at every position. Greedy output byte-identical to `STANDARD` on short prompts and on a 921-token prompt spanning four chunks and crossing the 512-position window (that run: 6.82 s → 0.92 s end to end). `Gemma4CpuGpuParityAccelTest` unchanged on both representations: relL2 7.42e-05 and 2.71e-06 against bounds 2e-03 and 1e-04, 0/63 argmax disagreements. 606 Class A tests pass.

That comparison found two defects a throughput number would not have: decode layer 0 consumed its activation by the unnamed form (right in the single-token plan, wrong here — **an activation of exactly zero out of every decode layer, with the model still emitting fluent tokens**), and the family routed to the interleaved generation loop regardless of phase strategy, so the batched plan was built and never driven.

### Stated plainly

- **Numerically this is not the decode path.** GEMM operands are FP16 with FP32 accumulation, and the two per-layer-embedding projections (F32 in this file) are narrowed to FP16 once at construction so they can be a GEMM's B operand at all. Prefill-only; `STANDARD` is untouched and its parity metrics are identical.
- **Q4_0 is refused by name**, not served wrong: `gemma4 batched prefill is Q8_0-only for the trunk; blk.0.attn_q is Q4_0`. Its `STANDARD` path is unaffected.
- **Decode in this plan costs 147.08 t/s against `STANDARD`'s 201.24.** The plan is opt-in (`--with-prefill-decode --batch-prefill-size N`), so nothing regresses unless asked for, but the trade is real and the cause is not yet established — the next measurement is whether the decode graphs' second copy of the weights is what costs it.
- **No new dispatch classes.** An earlier shape of this change added four and `DispatchLedgerTest` refused it ("this tree is collapsed; nothing may add to it"). The batched layers now sit in `layers/` beside `Qwen35BatchPrefillLayers`, the decode variation is a constructor flag, and the batched components are methods on the existing class. The ledger stays at `{fp16=30, q8_0=32}`.
- **0.6× prefill is still far off.** This closes the structural gap; #163's ladder — paired loads, the scored-attention scratch, dequantize-then-GEMM, the tiled B-stage, tensor-core attention — is what would be climbed from here, and each of those is now applicable where it previously was not.

https://claude.ai/code/session_014xVfr1SjQPTruQos3HArW9

mikepapadim:
## Conclusion: Gemma 4 prefill, measured end to end

Final evaluation on an **idle GPU**, reference re-measured in the same session, everything below with `--cuda-graphs` on both sides (see the correction at the end for why that qualifier matters).

RTX 4090 · TornadoVM 6.1.0-cuda (`3c6c5f12c`) · JDK 21.0.2 · driver 610.57.04 · CUDA 12.6 · `llama-bench` f172be756

### Q8_0 — `gemma-4-E2B-it-Q8_0.gguf`

| test | single-token | batched (b=512) | llama.cpp | batched ratio |
|---|---:|---:|---:|---:|
| pp512 | 196.53 | **7764.35** | 23752.94 ± 10.7% | **0.327** |
| pp512 @ d2048 | 184.91 | **3466.59** | 21724.37 ± 6.7% | 0.160 |
| tg128 | 200.42 | 197.40 | 236.77 | 0.834 |
| tg128 @ d2048 | 186.46 | 182.41 | 231.65 | 0.787 |

### Q4_0 — `gemma-4-E2B-it-Q4_0.gguf`

| test | single-token | batched (b=512) | llama.cpp | batched ratio |
|---|---:|---:|---:|---:|
| pp512 | 240.47 | **7289.80** | 24934.03 ± 7.5% | **0.292** |
| pp512 @ d2048 | 223.67 | **3433.84** | 22494.04 ± 6.5% | 0.153 |
| tg128 | 245.86 | 240.59 | 324.19 | 0.742 |
| tg128 @ d2048 | 226.82 | 220.02 | 314.99 | 0.698 |

Prefill is **39.5×** on Q8_0 and **30.3×** on Q4_0 against this family's own single-token path. The pp512 reference spreads exceed the 10% gate on Q8_0, so that ratio carries the reference's uncertainty; the d2048 references pass.

### How it got there

| step | what it was | pp512 |
|---|---|---:|
| — | decode loop run once per prompt token | 145 |
| `3ef3741c` | a batched prefill plan at all | 2767 |
| `a8a3706b` | attention keys staged through a transposed shared tile | 3153 |
| `fb2b002f` | depth split across blocks for the two narrow projections | 3797 |
| `bd490dbb` | gate/up decoded to FP16 once, then one FP16 GEMM | 4575 |
| `e4a59399` | query/key/value decoded the same way | 4771 |
| `201c9ec1` | FP16 split-K, so the last two got both fixes | 5857 |
| `aaf67b7e` | attention key tile halved — occupancy, not coalescing | 6469 |
| `79a57fdf` | per-layer embedding rows gathered across tokens | 7510 |
| `d0252706` | the same path for Q4_0 (18- and 20-byte decoders) | — |

Only the first is a port of #163's *structure*. The rest came from measuring this family rather than transplanting that one's ladder — and twice the measurement said something other than what the ladder predicted:

- **`aaf67b7e`** was going to be #163's warp-per-dot-product attention rewrite. Nsight said the kernel was neither bandwidth-bound (DRAM 0.70%, L2 hit rate 99.55%) nor compute-bound, but occupancy-bound at 41.7% *"limited by the required amount of shared memory"* — by the key tile added two commits earlier. Halving it took theoretical occupancy to 75% and the kernel from 776.6 to 551.3 µs. One constant, +9.5%, bit-identical.
- **`79a57fdf`** was not a kernel. The profile said 230 ms of kernel time per four chunks; the wall clock said 98 ms per chunk. The gap was the host-side per-layer embedding gather — 31.77 ms per chunk on Q4_0, counted by no profiler here.

Reading the emitted CUDA (`bd490dbb`) was worth more than any profile: `gemmMMAGateUpQ8` is **8,666 lines carrying 16 `mma.sync`**, against 1,179 for the FP16 `gemmMMA`. The Q8_0 GEMMs convert float→half in software because TornadoVM reaches the hardware conversion only through a store to a half array — filed upstream as beehive-lab/TornadoVM#1096 and #1097, with minimal reproducers; #1097 also confirms #1089 is still present at 6.1.0.

### Validation

- **Teacher-forced NLL against the host CPU**, which shares no kernel with any of this (`0dbfe7f9`). Prefix 320 at chunk width 256, so it crosses a boundary. Pooled: Q8_0 host 6.58276 → batched 6.58219 (**−0.00057** nats/token); Q4_0 6.31673 → 6.30038 (**−0.01635**). Favourable on both, per-passage signs mixed — arithmetic noise, not drift. Two builds on fixed inputs, not a quality benchmark.
- **Needle-in-haystack across chunk boundaries.** A fact at ~token 250 of a ~1500-token prompt, recalled at the end, correct on both fixtures at widths 256 and 512 and on the single-token path — six for six. Gemma 4's sliding-window layers (window 512) cannot see that far, so this is the 1-in-5 global layers, and it shows the KV cache and the SWA/global split both survive chunking.
- **Short prompts** (1–4 tokens, so a chunk that is almost all padding rows) answer correctly on both fixtures.
- `Gemma4CpuGpuParityAccelTest` and `Gemma4Q4_0ParityAccelTest` unchanged, bounds not touched; 606 Class A tests pass.

Two latent defects were found by comparing activations rather than by any benchmark, both of which produce *fluent, wrong* output: decode layer 0 consuming its activation by the unnamed form (an activation of exactly zero out of every decode layer), and the family ignoring the phase strategy so the batched plan was built and never driven.

### Corrections to my own earlier numbers in this PR

1. **`3ef3741c` says decode in the batched plan costs 147.08 against the single-token path's 201.24.** That compared a batched number measured *without* `--cuda-graphs` to a single-token number measured *with* it. Measured properly, with graphs on both sides: **197.40 against 200.42 on Q8_0 (−1.5%) and 240.59 against 245.86 on Q4_0 (−2.1%)**. The batched plan does not meaningfully cost decode.
2. **The decode campaign's figures earlier in this PR were measured with `--cuda-graphs`; the prefill campaign's were not.** Same commit, same GPU: tg128 is 133.75 without and 176.39 with. HEAD reproduces the decode campaign's 201.24/185.58 exactly (200.42/186.46) under that flag, and `717881b3` — the commit before any prefill work — measures identically to HEAD at 147.96 against 147.97 without it. **Decode has not regressed**; the two campaigns were reported in different configurations, which the measurement protocol warns about and I did not apply to myself.
3. One `llama-bench` run during this evaluation was taken while a jllm benchmark was still running and reported tg128 95.56 against the true 236.77. Discarded and re-run on a verified-idle GPU; every number above is from the clean run.

### What is not done

- **Prefill is 0.29–0.33 of the reference, not the 0.6 that was the target.** The remaining window is 258 ms of kernel time per four chunks, of which attention is ~35%; #163's tensor-core attention is the next step and was not reached.
- **pp512 @ d2048 is 0.15–0.16**, roughly half the d0 ratio — attention grows with depth and the current kernel is three passes with a global score scratch.
- **FP16/BF16 has no batched components**; the path is Q8_0/Q4_0/Q4_1 only, refused by name otherwise.
- The batched plan is **opt-in** (`--with-prefill-decode --batch-prefill-size N`). Nothing regresses unless it is asked for, and nothing improves either.
- CI rows, `models-and-backends.md` and `verification.md` are still unwritten for this family.

https://claude.ai/code/session_014xVfr1SjQPTruQos3HArW9

